PR #3, opened 2019-05-24
Add title and reorder menu

donho:
The new system will use the `weight` to order the menu.

### DIFF
--- a/content/auto-completion/_index.en.md
+++ b/content/auto-completion/_index.en.md
@@ -1,0 +1,4 @@
+---
+title: Auto-completion
+weight: 70
+---
--- a/content/binary-translation/_index.en.md
+++ b/content/binary-translation/_index.en.md
@@ -1,0 +1,4 @@
+---
+title: Binary Translation
+weight: 140
+---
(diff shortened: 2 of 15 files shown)


MetaChuh:
as we don't have a hugo or jeckyll engine preview yet, it would be unwise to judge the outcome, without being able to view it, except by predicting it, simulating the theoretical order outcome, based on it's weight.

@milipili 
please contact me in private, as i will not be able to make a valid review without your help.

milipili:
This is designed for gohugo. Temporary, you can take any hugo template locally. Sorry I am quite busy at work recently with some major internal release, but soon we will have a deployment per merge request as it should be.

donho:
@MetaChuh So can we still consider it as non-blockrd point ?
I don't really think these modifications are big deal. It'll be great if we can move forward.

milipili:
LGTM

MetaChuh:
@milipili 

splendid that it looks good to you 😉

i need your help to see the same on local hugo (55.6)
do you have some template files and hugo command line to render the md structure for me ?

they are not within this repo yet, and every member will need to be able to render every pr, in order to be able to validate each pr.

many thanks.

PR #11, opened 2019-07-12
Update README

tgburgin:
A couple of changes to improve readability. Bit concerned that we're linking to CONTRIBUTION.md and it's blank!


donho:
>  Bit concerned that we're linking to CONTRIBUTION.md and it's blank!

Yes, indeed. There should be a set of rules for the contribution. We (at least I) don't have yet any rules in mind. So I guess we e will see when the contributors are coming. Feel free to do your PR on this file if you have some good points.

PR #15, opened 2019-08-05
Augment 'searching' documentation.

sasumner:


sasumner:
Peter, I did a first-pass edit/augmentation of the top of the document.  I have more to do on the remainder of the content, but I thought I would get an PR in with the initial changes.  Plus it gives you the "flavor" of what I'd do to the document.  See what you think, tell me to go to Hell if you want. :)  Like I said, it is our document but it is really YOUR document!  :)

sasumner:
Peter, also, I don't want to mess with your "structure" of the document.  So maybe I put some stuff in there that I felt should be captured, but maybe not in the best place "structure-wise" (although I tried to pick the best place).  What I'd suggest for that is, accept it and move it around later when you make further edits for other reasons.  Is that reasonable?

PR #16, opened 2019-08-05
Add special find macro encoding info section to Searching

sasumner:
I felt the find macro stuff was different enough from normal macro stuff to put it in the "Searching" section rather than the normal "Macro" section.  

Plus I wanted to do this because I recently added the "Clear All Marks" command (1633) into the macro interpreting code and this completes that effort by documenting it.  :)  See issue # 1948 in N++ codebase.  

This is just a slightly reworked version of the documentation for this that was on the old Wiki page.  Corrected some errors, provided more verbose examples.

BTW, did I mess up line-endings with this or my earlier commit today, or this one?  I just noticed this output from my "git add" command:

    warning: LF will be replaced by CRLF in content/searching/index.md.
    The file will have its original line endings in your working directory

### DIFF
--- a/content/searching/index.md
+++ b/content/searching/index.md
@@ -69,7 +69,7 @@ The **Find what** and **Replace with** edit boxes have a dropdown arrow which al
 
 The **In selection** option will automatically be chosen by Notepad++ if a Find dialog window is opened when more than 1024 characters occur in the active selection.  The selected text will also be placed in the **Find what** box.  Running a **Count** or **Replace All** action without making other changes to the search parameters will result in *Count: 1 match* or *Replace All: 1 occurrence was replaced*, respectively, which is likely not what was intended.  The proper resolution for this situation is to change the **Find what** text if the intention is to search within-selection, or deselect **In selection** if the intention is to search for a fairly long block of text.
 
-The status bar area of the Find dialog keeps the user informed of what occurred during an action.  For example, it might say *Mark: 1 match* or *FInd: Invalid regular expression*.  Colors are used in the status bar for emphasis:  red for some sort of error; green or blue for various success or general information.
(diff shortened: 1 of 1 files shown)


content/searching/index.md
```diff
@@ -783,3 +783,107 @@ Now let's add some commenting
...
+
+The macro recorder only records when an actual Find family action (e.g. **Replace**, **Find All in Current Document**, etc.) occurs.  Thus you can tweak a future action's parameters (e.g. **Match case**, **Wrap around**, etc.) all you'd like, and all of that fiddling doesn't get remembered.  At the point where you peform an action, then a snapshot is taken of all of the parameters and the action, and this is logged in the macro memory as a proper macro sequence.
+
+While the user can simply record and use Find family macros, one can also edit those macros later to change or add to their functionality, so it is helpful to know the details of the macro sequences that were previously recorded.  While the details of how macros in general are recorded and stored in *shortcuts.xml* is discussed elsewhere, here are the details of what happens when Notepad++ saves a recorded Find family macro:
```
pryrt: I'd suggest changing `discussed elsewhere` to `[discussed elsewhere](../macros/)`: I've been using cross-document links when I do my edits, and I think it makes sense to have one here, too.

Actually, _is_ it really discussed on the Macros page?  If I'm reading correctly, not yet.  Probably need to add it to our TODO list to bring in the old macros-nitty-gritty from the Wiki to the Macro page.

sasumner: Okay, I'll update it with the cross-link when I next hit the document (I have more to do), since you are so "easy" on me.  

I didn't do it because the macros document is basically empty, so not much to jump to.  In this early stage of documentation revision it is a bit hard to do good cross linking; things are evolving.

Yes, I *presumed* that it will be discussed on the upcoming Macros pages.

PR #17, opened 2019-08-06
auto-completion fixes

pryrt:
content\auto-completion\:
* re-add note that the Delimiter setting might not do what you expect
* change "all the extra parameters" to "all the extra characters" when talking about the Delimiter setting.
* in the sorting section, remove reference to "a tool will be made available", since it likely never will be, at this point
* in the sorting section, remove reference to "textfx plugin does, which is installed by default", since it's not installed (or recommended)
* merge the two #sorting sections, since they had the same basic content, worded differently (they had come from two different files on the original Wiki pages)

content\preferences\#delimiter:
* add the note about the contents of this field
* add the note about it not working as you might expect for auto-completion

these fixes were noticed as needed while I was helping with: https://notepad-plus-plus.org/community/topic/18053/custom-static-function-with-parameter-hints


pryrt:
> Spelling in one place: "additinal" should obviously be "additional".

Ooops. :-)

> I can't agree or disagree with info about the older N++ version references; how far back should this new documentation attempt to reference? Hmm.

Normally, I would not have referenced that far back.  Basically, I was re-introducing a comment that was present back in the NppWiki++, which said the feature had been introduced but was still buggy.  Apparently, it's still buggy.

> Semi-related commentary: The ability to add to the "Word character list" was done "recently" but with a very specific goal in mind. Perhaps it is unfortunate that users will expect other features that use word characters/boundaries to also respect the custom list? 

Well, the NppWiki++ specifically applied the "Word character list" to Smart Highlighting and to Auto-completion, so those are probably considered the "canonical" uses for the list.

> Maybe issues should be opened as these things are discovered, and all possible relevant features should be made to use the custom list? 

For auto-completion, all I know is that the `:` didn't work for recognition of the keyword for applying parameter hints.  But I'm not sure that punctuation was really ever meant to be allowable in the word character list.

> (Regex searching also has the concept of word characters/boundaries; changing how that works to respect this list would be interesting, so I hope users don't think of this!)

I think I will go specifically disallow regular expression "words" from the Delimiter setting.

pryrt:
> Just asking: Does Unicode have any influence on the default word characters? (BTW, you are a far better Unicode knowledgeable person than I am)

I had thought it might, but hadn't tried.  
In terms of Smart Highlighting, it definitely accepts anything that Unicode considers a "word" character (alphanumeric and underscore), so letters (in any script), digits, letterlike symbols, and the superscript digits or enclosed/circled digits.  But punctuation (except for underscore) is not considered a word character, nor are mathematical operators, symbols, emoji, arrows, etc.

I also tried out the additional Delimiters: in Smart Matching, the generic punctuation I tried (`.:,;-`) seem to work, so I will delete the note in the preferences section.

I'm thinking for clarity for now, since I don't have time to do up a custom auto-completion test-bed, that I will comment out the mention of Delimiters for auto-completion, and thus not need the version-note there, either.

Another commit coming soon

donho:
> Normally, I would not have referenced that far back. Basically, I was re-introducing a comment that was present back in the NppWiki++, which said the feature had been introduced but was still buggy. Apparently, it's still buggy.

For me the document should present the feature of application, and only that. For the remain stuffs, we've got issues tracker on GitHub -except the bug is declared as feature officially. :)

PR #28, opened 2019-09-17
Update searching document, primarily Find-result window area

sasumner:
I had hoped to do more, but I think I got some good info into these changes.  I didn't update for the new contribution guidelines decisions, mainly because I didn't touch any areas of the existing document where this came into play (AFAIK).  I can do that for the entire document in a subsequent edit (still much more to do).

### DIFF
--- a/content/docs/searching.md
+++ b/content/docs/searching.md
@@ -11,9 +11,15 @@ All keyboard shortcuts mentioned below are the default values, but are configura
 
 ## Dialog-based Searching
 
-The most powerful set of searching features is found in the standard dialog-based Find / Replace / Find In Files / Mark dialog.  This dialog has one tab for each of the forementioned searching-related features.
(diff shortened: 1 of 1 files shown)


donho:
@sasumner You can merge it directly into master please.
If you don't want to, @pryrt could help.

sasumner:
@donho 

>merge it directly into master

Yes, I will do so.  I have no reluctance to do my own merging of _documentation_ into "master", I just don't want to do my own merging of _code_.  :)

sasumner:
@pryrt 

>Looks good. Great detail.

Glad you like it.  I'm attempting to pull together all the little scraps of notes I've taken over the years on the minutiae of how all this stuff works.  Even after figuring it out long ago, I still can't remember how stuff works sometimes when I want to use it (and then I refer to my note scraps) but in the future I will consult the new docs!  

The "searching" area is of particular interest to me.  Much more to do!

BTW, my goal is to just get the stuff in there, initially.  It may not be "organizationally" the best, but of course I attempt to do a good job with that.  If you want to take an organizational-hand and move the content around, and feel that you have a knack for that kind of thing, please, Please, PLEASE...do it. :)

PR #34, opened 2019-10-06
Plugin Communication (was: Messages and Notifications)

pryrt:
Convert the Messages and Notifications page from the old NpWiki++


pryrt:
added.  thanks for the feedback.

README.md
```diff
@@ -24,6 +24,7 @@ Summary
 * [Plugins](content/docs/plugins.md)
 * [Preferences](content/docs/preferences.md)
 * [Themes](content/docs/themes.md)
+* [Plugin Development: Messages and Notifications](content/docs/messages-notifications.md)
```
donho: 1. The short title "Plugin Development" is enough IMO
2. "Plugins" is better to be followed by "Plugin Development" IMO

content/docs/messages-notifications.md
```diff
@@ -0,0 +1,1692 @@
+---
+title: Plugin Development: Messages and Notifications
```
donho: The short title "Plugin Development" is enough

pryrt:
Okay, I've reordered so it's after "Plugins".  I've renamed to "Plugin Communication", which still focuses on the "plugin" aspect, is reasonably short, but limits the scope of the content to communication (which is handled through the messages and notifications described).

donho:
"Plugin Communication" is a well found one.

PR #36, opened 2019-10-06
Config file syntax

pryrt:
copy over the contents from the old NppWiki++ page about the configuration file syntax.  Clean up to match the style of the npp-usermanual

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ Summary
 * [Plugins](content/docs/plugins.md)
 * [Preferences](content/docs/preferences.md)
 * [Themes](content/docs/themes.md)
+* [Syntax for the XML-based Configuration Files](content/docs/config-syntax.md)
 * [Command Line Arguments](content/docs/command-prompt.md)
 * [Shell Extension](content/docs/shell-extension.md)
 * [Binary Translation](content/docs/binary-translation.md)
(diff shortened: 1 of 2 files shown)


pryrt:
I've updated the title; if it's still not explicit enough, please offer a better suggestion.  Thanks.

donho:
> copy over the contents from the old NppWiki++ page about the configuration file syntax. Clean up to match the style of the npp-usermanual

I read it again and I'm wondering if it's really needed - NppWiki++ contains a lot of useful information, but also a lot of redundant info.

The goal of npp-usermanual is not restore NppWiki++, but provide the simple and direct information to users.

IMO, npp-usermanual can do without this part. In `config.xml` the node naming is quit self-explained. And users don't need us to explain what it is xml files - there are a bunch of sites do better job than us. If you think it's really necessary, I'd suggest to write it from scratch, with the following structure:
```
configurations files
- config.xml
- langs.xml
- stylers.xml
- shortcuts.xml
- contextMenu.xml
```

pryrt:
I agree that `contextMenu.xml` is the most in need of description.  And yes, the descriptions of XML are probably extraneous to this document.

I think power users would like a reference here for _some_ of the other details, but as it is, it is probably overboard.

I'll spend some effort to clean it up, and bring it down to a more manageable size, so I don't overwhelm the normal user, but still give power-users enough to dig into the details.

pryrt:
Okay, I pared down the Config File Syntax from about 1200 wordy lines to a much more manageable 200 lines.  It focuses on what's readily customized (and emphasizes using the GUI for settings when possible), but gives enough information to get power-users started down their super-customization paths.

PR #65, opened 2019-12-19
Add <GUIConfigs> commandLineInterpreter setting

xylographe:
Only a draft ATM, because the version number of the next NPP release is yet unknown.

Documents new ‘hidden’ `config.xml` setting, added in commit  https://github.com/notepad-plus-plus/notepad-plus-plus/commit/4228c315fd53298dad0712a0bce74452fd69994b.


xylographe:
Draft status removed: Don has confirmed that [the next release will be v7.8.3](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/2877#issuecomment-569422932).

PR #71, opened 2020-01-15
Suggest to modify this description

zhangyiqiu:
Current description may have issues.

Generally speaking, "chinese" refers to the official language of China, but now, in this usermanual, use "Chinese" refer to the language used in Taiwan. So it is recommended to stay neutral, changed to use the full description.

### DIFF
--- a/content/docs/binary-translation.md
+++ b/content/docs/binary-translation.md
@@ -26,7 +26,11 @@ For modifying an existing xml files, here is the way to visualize your modificat
 <div>
 
 <div align="center" style="margin: 20px; display: inline-block;">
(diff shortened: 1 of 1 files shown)


donho:
Remove "province" term from your description. I will see what can I do with your PR.

zhangyiqiu:
Edit completed

donho:
Before the invention of "Chinese Simplified", there was only "Chinese", ie. the so called "Chinese traditional" of nowadays.
Here's the question: Why do we need to rename "Chinese" to "Chinese traditional" ? For me it's misleading. 
I don't really care to "aligned" with the other software by using this wrong term. Let's make it right in Notepad++ : there are "Chinese" & "Chinese simplified", period.

PR #72, opened 2020-01-18
UDL info table -- fix rendering

pryrt:


donho:
@pryrt  If you have tested and it works, please merge it directly.
Currently I fixed it only on the web by using table tag directly.

PR #75, opened 2020-02-23
Add info about controlling N++ sound via Sound Mixer (due to N++ PR 7951)

sasumner:
This is a follow-up to functionality change introduced by https://github.com/notepad-plus-plus/notepad-plus-plus/pull/7951 originating from issue https://github.com/notepad-plus-plus/notepad-plus-plus/issues/7950

-----------

This PR is for a "basic" change to the docs.

Don requested an "extended" change which includes screenshots.  This can be done, but it felt rather "heavy" to do it inline where the "basic" change was made.  Perhaps this is best added in an "appendix" area?  **Looking for suggestions on how best to do it.**  I'll follow up with a future PR that provides an agreed-upon "extended" change.

Here are the screenshots I'd use, in the order I'd put them (of course additional text describing the screenshots and the process would be added):

image

image

image


pryrt:
Hmm... With the screenshots, it's feeling like the Cheat Sheet in the Other Resources page... But I am reluctant to make a super-crowded image rich page.

I wonder if that single page should be turned into a How To folder, with one page per topic (so 3 separate pages for the notepad.exe replacement, tab navigation, and new volume control)

sasumner:
I merged the "basic" change, and I will keep thinking about the longer version of documenting this.

pryrt:
@sasumner ,

[In the forum, Alan discovered](https://community.notepad-plus-plus.org/post/51665) that the Notepad++ entry doesn't show in the Volume Mixer until _after_ the first unsuccessful search -- so people who want to turn it off before getting beeped at cannot find it.  I don't know if that's something that should be changed in the docs to match reality, or if there's a way to improve the feature in code, so it matches the expectation to be able to do it at any time, not just after a beep has already occurred.

Dosahka:
it seems that this still not fixed properly? i have 7.8,6 x64 and this trickery to mute the failed search notif sound in N++ is not working (N++ not appearing in volume mixer), multiple pc restart and N++ reinstall. At this point i think i would much appreciate if this can be turned off in setting with a tickbox. Simple and neat.

sasumner:
>so it matches the expectation to be able to do it at any time, not just after a beep has already occurred.

Why would anyone be bothered by it if they haven't heard it yet? :-)

I will point out that there has NEVER been a way to turn that beep off, until I added one.

jpvanleeuwen:
> > so it matches the expectation to be able to do it at any time, not just after a beep has already occurred.
> 
> Why would anyone be bothered by it if they haven't heard it yet? :-)
> 
> I will point out that there has NEVER been a way to turn that beep off, until I added one.

Having a possibility to turn the sound off is a blessing, so thanks again for providing the possibility.

To further improve the user xp maybe it would be better if the notepad app is always in the volume mixer (if that is possible). I have now muted the Notepad sound but when I want to turn it on again (which I don't, but another user might) I have to first do an unsuccessful search to get notepad in the volume mixer and then unmute it. This is a bit ackward isn't it? Should at least be documented as such.

But again, rather have it this way than not being able to disable it at all.

jpvanleeuwen:
> it seems that this still not fixed properly? i have 7.8,6 x64 and this trickery to mute the failed search notif sound in N++ is not working (N++ not appearing in volume mixer), multiple pc restart and N++ reinstall. At this point i think i would much appreciate if this can be turned off in settings with a tickbox. Simple and neat.

I cannot reproduce this on Window 10 with version 7.8.6. x64. At my system the Notepad++ app shows up after unsuccessful search.

Notepad++ v7.8.6   (64-bit)
Build time : Apr 21 2020 - 15:23:03
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS Name : Windows 10 Home (64-bit) 
OS Version : 1903
OS Build : 18362.778
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll

Dosahka:
> @Dosahka
> 
> Are you performing an unsuccessful search first? (the N++ only appears in the volume mixer after an unsuccessful search, as it appears).

as i mentioned "this trickery to mute the failed search notif sound in N++ is not working (N++ not appearing in volume mixer)" trickery = workaround (or only solution that we have now) yes i have tried that following the steps as described.

I would like to point out that for me it runs in administrator mode by default (UAC is set to the lowest) and only sound output i have is my gaming headset with Logi G HUB, so not the native driver on my motherboard.

Also i'm happy to take this conversation to a private conversation on Discord to leave this thread tidy.

Debug info just in case:
Notepad++ v7.8.6   (64-bit)
Build time : Apr 21 2020 - 15:23:03
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : ON
Local Conf mode : OFF
OS Name : Windows 10 Pro (64-bit) 
OS Version : 1909
OS Build : 18363.778
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll NPPJSONViewer.dll

jpvanleeuwen:
@Dosahka 
> Also i'm happy to take this conversation to a private conversation on Discord to leave this thread tidy.
DiscordTag?

Dosahka:
> @Dosahka
> 
> > Also i'm happy to take this conversation to a private conversation on Discord to leave this thread tidy.
> > DiscordTag?

Dosi#2866

sasumner:
@jpvanleeuwen 

> it would be better if the notepad app is always in the volume mixer (if that is possible)

It is my impression that applications don't get recognized by the sound mixer until they attempt to make a sound via the API that the sound mixer recognizes.

jpvanleeuwen:
@sasummer
Yes, or change the volume I read somewhere. Changing the volume and setting it back to original on startup would be a workaround to always show the N++ app in the volume mixer.
I.m.h.o. having a sound section in the Preferences with check boxes for all possible sounds would be a better solution than the current solution. Would you consider it?
Again, better the current solution than no solution.

Andargor:
> @sasummer
> I.m.h.o. having a sound section in the Preferences with check boxes for all possible sounds would be a better solution than the current solution. Would you consider it?

I know it has been said that the been was always there, but I've never heard it until 7.8.6, and N++ doesn't appear in my mixer. Supposedly, you need to have the device set to "allow exclusive control" by applications, but for me that is a no go since I have some apps that misbehave and mess with my volume.

So I second the motion to have a setting in Preferences to turn off sound, it's what I search for first in vain before realizing that I had to use a convoluted but futile procedure in Windows.

Andargor:
This worked for me, an an elevated CMD prompt (run as admin). This will permanently disable the beep for the system.

```dos
C:\WINDOWS\system32>net stop beep
The Beep service is stopping........
The Beep service could not be stopped.


C:\WINDOWS\system32>sc config beep start= disabled
[SC] ChangeServiceConfig SUCCESS

C:\WINDOWS\system32>
```

HTH someone.

PR #77, opened 2020-04-06
Update preferences descriptions for missed v7.8.x changes

pryrt:
Update preferences descriptions for missed v7.8.x changes
* v7.8.1: add the MISC setting for Monospaced Find Font
* v7.8.1: move SQL-backslash option from MISC to Languages
* v7.8.3: add the MISC setting for Don't Fill Find Field
* v7.8.3: move the auto-indent option from MISC to Auto-completion
* Note that File Association changes must be made in Administrator mode
* Clarify "Make Language Menu Compact" per [this Community Post](https://community.notepad-plus-plus.org/post/51969)


content/docs/preferences.md
```diff
@@ -119,20 +119,27 @@ Select the name of the filetype in the left column; select one of the extensions
 
 To unregister an extension, click it in the right column, and use the left arrow `←` to remove it from the Registered Extensions column.
 
+For this section to work, please run Notepad++ in Administrator mode, since it is modifying the registry.
+
```
dail8859, line 123: I'd recommend putting this as the first sentence of this section, or at least adding a bold "**Note**" at the beginning of this line. Probably shouldn't assume users will read the entire section clear to the last line. So I'd suggest calling this part out in particular since this caveat of "this will only work under specific conditions" is important.

PR #78, opened 2020-04-08
clarify number of backreferences

pryrt:
Clarify that there _can_ be more than just 9 capture groups, as long as the right backreference syntax is used.  Make sure it's clear how to use ℕ>9 in both backreferences and substitutions.

In the forum, a user [recently claimed](https://community.notepad-plus-plus.org/post/52274) that only 9 groups are allowed! and I want to make sure that the documentation explicitly contradicts that fallacy.


pryrt:
need to get rid of conflicts; I'll be back soon

PR #79, opened 2020-04-08
clarify capture group backreferences with N>9 are possible

pryrt:
unbelievably, [recently someone asserted there were only 9 backreferences possible](https://community.notepad-plus-plus.org/post/52274)!

After re-reading the docs, I decided the docs needed to be more emphatic that more than 9 are possible with the `\g{N}` for syntax for backrefs and `${N}` syntax for capture groups in substitutions.

### DIFF
--- a/content/docs/searching.md
+++ b/content/docs/searching.md
@@ -462,24 +462,40 @@ Anchors match a zero-length position in the line, rather than a particular chara
 
 
 
-#### Groups
+#### Capture Groups and Backreferences
 
-* `(`_subset_`)` ⇒ Parentheses mark a _subset_ of the regular expression. The string matched by the contents of the parentheses (indicated by _subset_ in this example) can be re-used as a backreference or as part of a replace operation; see [Substitutions](#substitutions), below. Groups may be nested.
+* `(`_subset_`)` ⇒ _Numbered Capture Group_: Parentheses mark a _subset_ of the regular expression, also known as a _subset_ expression or capture group. The string matched by the contents of the parentheses (indicated by _subset_ in this example) can be re-used with a backreference or as part of a replace operation; see [Substitutions](#substitutions), below. Groups may be nested.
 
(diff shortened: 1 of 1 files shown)


pryrt:
I think that what you describe is already documentd in [Control Flow](https://npp-user-manual.org/docs/searching/#control-flow): `(?ℕ)`.  (I hate to admit, but I couldn't figure out what you were trying to describe until I read the docs for `(?ℕ)` -- even with the example, I couldn't tell that it was taking the _pattern_ from match one -- the "TEXT" wasn't clear enough to me -- and re-applying that _pattern_ in the new location).  It appears the [official boost](https://www.boost.org/doc/libs/1_70_0/libs/regex/doc/html/boost_regex/syntax/perl_syntax.html#boost_regex.syntax.perl_syntax.recursive_expressions) calls it "recursive", though it doesn't quite meet my standard thoughts on "recursive"

At one point, Guy did give me a whole bunch of changes that I implemented for him; he had fixed the examples (which were pre-Boost examples), among other things -- yeah, PR #59 .  So he has helped with the quality of this.

Anyway, I'll merge as-is; if you think  `(?ℕ)` needs more (or isn't what you were describing), feel free to submit your own PR. :-)

sasumner:
Actually, that other section you linked to covers it, but I may try to improve its verbage.  I didn't think to look for it under "control flow"; truly it seems to fit better "near" the backreferences section (since it is some sort of reference to a prior thing).  If I do decide to try to make it better, I'll leave it in control-flow and simply attempt to beef it up a bit.  The example given is rather weak because it isn't practical; no one would ever "subexpression-reference" something as simple as `\w` -- they'd just retype it.  But it can come in very handy with more complicated expressions that occur multiple times.  It may be tough to come up with a good example that is short enough to embed in the docs.

Sorry for my confusing explanation from before.

Guy would have the better opinion on the "recursive" nomenclature.  I think maybe these things have some uses when truly recursive regexes are written; such as nested parenthesis type problems.  But I'm out of my element as I truly don't understand those.

PR #81, opened 2020-04-15
Apply updates to preferences docs to add in the new v7.8.6 preferences

pryrt:
- Editing: Vertical Edge: multi-line mode
- Global Styles: unavailable colours selections are now greyed out

Not new to v7.8.6, but things that were unclear in preferences.md
- Auto-completion: _n_th character gets greyed out when parent option disabled
- Colour Selectors: right click to strike out the colour to indicate it should inherit from default settings

### DIFF
--- a/content/docs/preferences.md
+++ b/content/docs/preferences.md
@@ -54,7 +54,9 @@ These influence editing (carets, code-folding, line wrapping, and more).
     * `☐ Show vertical edge`: shows a vertical edge at the specified location, often used to indicate the right margin for manually setting the number of characters per line
     * `☐ Line mode`: the vertical edge is indicated by a solid vertical line, styled with the **Settings > Style Configurator > Global Styles > Edge Color: Forground colour**
     * `☐ Background mode`: the vertical edge is indicated by styling the _background_ of the text to the right of the edge with the colour from **Settings > Style Configurator > Global Styles > Edge Color: Forground colour**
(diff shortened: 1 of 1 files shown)


pryrt:
prompted by v7.8.6 RC announcement to get the new preferences documented; want to get this merged about the same time that v7.8.6 is made official, so that docs don't lag NPP application for too long.

pryrt:
Yeah, experimentally, v7.8.5 didn't grey it out, but v7.8.6 did.  For example, Caret Color:

**v7.8.5**:
image

**v7.8.6 rc**:
image

pryrt:
Thanks for approving.  I'll merge once v7.8.6 officially releases.

sasumner:
That is so awesome.  Before I'd be left wondering "do _both_ foreground and background really apply to this setting?", especially for those items where it is unclear which really is the foreground and which really is the background.

pryrt:
Weird.  Guy [just pointed out](https://community.notepad-plus-plus.org/post/52729) that with his v7.8.5, it does _start_ greyed out, but then if he changes to Bespin theme and back, then it's no longer greyed out.  For me, my normal portable never shows it greyed out, but a fresh unzip will start greyed out and Guy's sequence will make it un-grey.

It looks like there might be a bug in that only-one-color-applies-then-grey-out-other feature.  I can no longer guarantee that it always works in v7.8.6... I wonder if I should leave that description in?

PR #95, opened 2020-06-05
Add "Preferences for Advanced Users" section for non-UI options

donho:
And add "regexBackward4PowerUser"="yes" for enabling regex backward search.


donho:
@pryrt I saw there's another config-files section, where there's advanced settings `commandLineInterpreter`. I'll try to regroup them together in this PR.

PR #96, opened 2020-06-16
Fix Typo Manuel -> Manual

djwhitten:
This is spelled properly later in this document, so I don't know if this is intended to be a simple test of ability to supply a  properly formatted Pull Request or a real typo.
If it's real, here is a Pull Request to fix it.

### DIFF
--- a/CONTRIBUTION.md
+++ b/CONTRIBUTION.md
@@ -9,7 +9,7 @@ Your pull requests are welcome; however, they may not be accepted for various re
 ### Guidelines for pull requests
 
 0. KISS - Keep It Simple, Stupid.
-1. Respect Notepad++ User Manuel style.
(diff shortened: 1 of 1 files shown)


djwhitten:
Do I need to fill out any "contributor copyright document"  ? 
I expected that is why this CONTRIBUTIONS.md was referenced,
but it didn't say anything about it.

pryrt:
> Do I need to fill out any "contributor copyright document" ?

Never seen such a thing on the few open source projects I've contributed to -- but I don't have a wide experience.

We do probably need to define a license for the document, which would likely end up using one of the standard open-source documentation licenses.  As long as you're willing to have it be released under such a license (whatever we choose in #99), we can incorporate your changes.

PR #97, opened 2020-06-16
Provide URL for this manual if unknown to getting-started.md

djwhitten:
Many times, a printed manual does not include the URL to find it on the Internet.
I'm wondering if I should have said:

The most up-to-date version of this manual is available on the website https://npp-user-manual.org/

Anyway, I think it helps to give the URL.


pryrt:
I have vague recollections that I tried to add such a link when we were first porting this, and that others thought it wasn't necessary, but I cannot recall where that is, and haven't had a chance to dig into the PR/commit history to see if my recollection is right or not.

donho:
on then page https://npp-user-manual.org/ to mention https://npp-user-manual.org/ ?
I think it's just redundant.

Issue #113, opened 2020-09-05
Multiple typos in assertion syntax in (?(assertion)YesPattern|NoPattern)


Medvedishce:
See here: [Control flow](https://npp-user-manual.org/docs/searching/#control-flow)

Inside of _assertion_ syntax part of 
- `(?(assertion)YesPattern|NoPattern)` there are multiple typos.
Here are how it is (bold - typos):
  - `(ℕ)` ⇒ true if ℕth unnamed group was previously defined

  - `(`**`?`**`<name>)` or `(`**`?`**`'name')` ⇒ true if group called name was previously defined

  - `(?=lookahead)` ⇒ true if the lookahead expression matches

  - `(?!lookahead)` ⇒ true if the lookahead expression does not match

  - `(?<=lookbehind)` ⇒ true if the lookbehind expression matches

  - `(?<!lookbehind)` ⇒ true if the lookbehind expression does not match

  - **`(?`**`(R)`**`)`** ⇒ true if inside a recursion

  - **`(?`**`(Rℕ)` ⇒ true if in a recursion to subexpression numbered ℕ

  - **`(?`**`(R&name)`**`)`** ⇒ true if in a recursion to named subexpression name

- and how it must be:
[boost_regex.syntax.perl_syntax.conditional_expressions](https://www.boost.org/doc/libs/1_70_0/libs/regex/doc/html/boost_regex/syntax/perl_syntax.html#boost_regex.syntax.perl_syntax.conditional_expressions)
  - `(ℕ)` ⇒ true if ℕth unnamed group was previously defined

  - `(<name>)` or `('name')` ⇒ true if group called name was previously defined

  - `(?=lookahead)` ⇒ true if the lookahead expression matches

  - `(?!lookahead)` ⇒ true if the lookahead expression does not match

  - `(?<=lookbehind)` ⇒ true if the lookbehind expression matches

  - `(?<!lookbehind)` ⇒ true if the lookbehind expression does not match

  - `(R)` ⇒ true if inside a recursion

  - `(Rℕ)` ⇒ true if in a recursion to subexpression numbered ℕ

  - `(R&name)` ⇒ true if in a recursion to named subexpression name

P.S: 
All this typos were found when I painfully did example for #112 so all corrections tested before.
P.P.S.:
[...]

Issue #112, opened 2020-09-05
Improper placement of (?>pattern) in Assertions 


Medvedishce:
Hello!

Here we have false declaration:
Look here: [Assertions](https://npp-user-manual.org/docs/searching/#assertions)

1) Assertions stated as:

> These special groups consume no characters. Their successful matching counts, but when they are done, matching starts over where it left.

2) `(?>pattern)` declared inside Assertions but CONSUME characters and so can't be placed in Assertions chapter.

In the PERL documentation it is listed without grouping like this:

> **Independent sub-expressions**
> `(?>pattern)` _pattern_ is matched independently of the surrounding patterns, the expression will never backtrack into pattern. Independent sub-expressions are typically used to improve performance; only the best possible match for pattern will be considered, if this doesn't allow the expression as a whole to match then no match is found at all.
> 

Considering what it do `(?>pattern)` can be placed in **Search modifiers** or **Control flow** chapters.

My suggestion is to place it inside of **Control flow** chapter before `(?(assertion)YesPattern|NoPattern)` and add an optional guidelines to it like this:

- `(?>pattern)` => Independent sub-expression
Match _pattern_ independently of surrounding patterns. Search will never backtrack into independent sub-expression.
Independent sub-expressions are typically used to improve performance, only the best possible match for pattern will be considered, if this doesn't allow the expression as a whole to match then no match is found at all.
On other side you may be forced to use it, for example if later in expression will be back-reference or `(?(assertion)YesPattern|NoPattern)`:
  - the regex `(?:(100)|\d{3}) apples (?(1)YesPattern|NoPattern)` in plenty cases will find '100 apples NoPattern' because if _YesPattern_ faled, search will backtrack to the beginning and try next alternative and yes, '100' also 3-digits number.
  - the regex `(?>(100)|\d{3}) apples (?(1)YesPattern|NoPattern)` will only search '100 apples YesPattern' even if _YesPattern_ faled, and 'xxx apples NoPattern' only if xxx<>100.

Medvedishce:
Here is an example in details, but I don't want to place it inside declaration part of documentation, may be as some new example to demonstrate how to use **Conditional Expressions** and its tricks. 

**File1:**
100 apples in a box
100 apples in a barrel
500 apples in a barrel

**Search Results:**
`(?:(100)|\d{3}) apples (?(1)in a box|in a barrel)`
  File1  (3 hits)
	Line 1: 100 apples in a box
[...]

PR #114, opened 2020-09-06
Fix Independent sub-expression and Assertions

pryrt:
1. Per #112, moved Independent sub-expressions into the main control flow (since they do consume characters) and increased description to make their purpose more obvious
2. Per #113, made the Assertions section more consistent (stripped out the `(?...)` wrappers which were incorrectly present on some, but not all, of the list)


Medvedishce:
@pryrt , I took a look and generally have no problems with content but about lexis in one paragraph:

>  * The regex `(?:(100)|\d{3}) apples (?(1)YesPattern|NoPattern)`, because it does not use the independent sub-expression, in many cases will find `100 apples NoPattern`, even though you expected `YesPattern` to be used when because `100` was matched.  This happens because, if `YesPattern` failed, the search will backtrack to the beginning and try next alternative, where `100` matches `\d{3}`, but means that `?(1)` does _not_ match so the expression uses `NoPattern`.


I think there is excessive repetition of word "because".
And in "to be used **when** **because** `100` was matched" I think only one of this two needed.
May be you can rephrase paragraph?

PR #120, opened 2020-10-25
Add a notice for function list

donho:
Add a notice to not use RegEx look behind operations for function list.
Please check:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9008#issuecomment-715498220
for further information.


Uhf7 (commented):
They don't work, not doesn't.

Fell for the same thing in one of my issue titles recently, so I investigated ...

Issue #125, opened 2020-10-29
Docs for number captured groups incorrect


dail8859:
https://github.com/notepad-plus-plus/npp-usermanual/blob/master/content/docs/searching.md#capture-groups-and-backreferences

States that \gℕ, \g{ℕ}, \g<ℕ>, \g'ℕ' should be the correct way to reference capture groups > 10 however it did not work for me. 

Referencing https://github.com/notepad-plus-plus/notepad-plus-plus/issues/2514 shows $N does in fact work. It is also documented in https://github.com/notepad-plus-plus/npp-usermanual/blob/master/content/docs/searching.md#substitution-escape-sequences that $ℕ, ${ℕ}, \ℕ ⇒ Returns what matched the ℕth subexpression

pryrt:
\gℕ is not a substitution escape: it is a backreference, used in the MATCH (FIND WHAT) portion of the regex; it is not used in the replacement/substitution.  That's why it's not documented in the substitution section.

With the text
```
abcdefghijklmnopqrstuvwxyzm
```
The FIND expression `(.)(.)(.)(.)(.)(.)(.)(.)(.)(.)(.)(.)(.)(.).*\K\g13` will match the m at the end of the line, because `\g13` is a backreference to the 13th matched group.

Do we need to make it more clear in the backreference section that backreference notation is solely for the match/find portion, and not meant for the replacement syntax?  Or would you rather have that note in the substitution?  Or both?

dail8859:
Oooh...yeah you are totally correct. I got those two confused (and didn't read enough of the surrounding documentation) so completely an error on my part :)

pryrt:
If an experienced user like you gets confused, I'm betting that newbies do, too.  I might try to come up with a way to make that more clear in both the search and the replace sections...

dail8859:
The problem was I knew there was something special about groups > 10 so as soon as I found something that mentioned groups > 10 I thought that was the correct information.

pryrt:
I've got a version with some clarifications in both the backreference and replacement sections
* https://github.com/pryrt/npp-usermanual/blob/searching-backref/content/docs/searching.md#capture-groups-and-backreferences
* https://github.com/pryrt/npp-usermanual/blob/searching-backref/content/docs/searching.md#substitution-escape-sequences

Do you think that would make it better for newbies (or confused old-timers)?

PR #127, opened 2020-10-31
backreference vs substitution clarification

pryrt:
Since both @dail8859's #125 and the forum [question](https://community.notepad-plus-plus.org/topic/20219/replace-character-in-capture-group) came up the same week, obviously the docs need clarifying.  Here's my attempt.


donho:
@dail8859 
Could you approve this PR please?

dail8859 (approved):

dail8859:
@donho Got busy and this fell off my radar. Apologize for that.

@pryrt Thanks for keeping the documentation maintained even when I screw up :)

donho:
@dail8859 
No problem at all! Thank you for reviewing.

PR #128, opened 2020-11-01
Add new function list info into config-files.md

donho:
Add new config info regarding function list.

### DIFF
--- a/content/docs/config-files.md
+++ b/content/docs/config-files.md
@@ -205,20 +205,20 @@ From v7.9.1 the file structure of function list parsers have been changed, so pl
 The `functionList` folder contains a separate XML file (function list parse rule) for each language's function list capability.
 Each function list parse rule links to a language with the language default name. For example the file name of php language parse rule is `php.xml`, the file name of Java language parse rule is `java.xml`, Check [overrideMap.xml](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/installer/functionList/overrideMap.xml) for the naming list of all supported programming languages.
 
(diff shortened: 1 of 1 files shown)


content/docs/config-files.md
```diff
@@ -195,9 +195,32 @@ The `<WordsStyle>` `colorStyle` attribute decides whether to use the defined col
...
+
+* Define your UDL example
+```
+<association id= "krl.xml"				userDefinedLangName="KRL"/>
+```
+Here you define a parser rule file name for your KRL UDL. While you open a file which is recognized as KRL file, then function list engin will load `functionList\krl.xml` to show the KRL function list. If you have no KRL UDL defined in your Notepad++, you have to define a dummy one (with the name "KRL") to make it work.
+
+### v7.9 and previous versions
 
 If you want to add **Function List** capability for your User Defined Language (UDL), you can.  You need to add two groups of information:
```
pryrt, line 225: Because we've split this up, the v7.9-and-earlier section no longer mentions the name of the config file.  We should insert a sentence at the beginning of this line:

> The `functionList.xml` config file contains XML entries for each language's function list definition, as well as a map that tells Notepad++ which section of the XML is applied to each file type.

pryrt (changes requested):
After adding the one sentence I mentioned, it looks good (and better than my original submission).

Feel free to merge after the one sentence has been added.

Thank you.

content/docs/config-files.md
donho: @pryrt 
You're right about it. I will add it.

PR #134, opened 2020-11-18
describe Find dialog behavior multiple monitors

wistlo:
Describes the behavior of Find dialog across multiple monitors and method to bring Find dialog to instance with current focus.


wistlo (commented):
[Ugh, need to fix the wording, I can't edit/delete:]

...will open on the same monitor used by the previous invocation regardless of which monitor currently shows the active Notepad++ instance.

PR #150, opened 2021-01-12
update config-files: macros

### LINKED ISSUE #149: config#macro needs improvement
Regarding the paragraph has two issues:
> You can use any Scintilla or Windows message that does not return a value, that passes an integer in wParam, and either an integer or string in lParam. There are some messages that require strings in the wParam, or various data structures: those will not work in a macro.

1. The first should probably just mention Scintilla messages; most standard Windows messages don't apply, or otherwise aren't reasonable from the macro environment.
2. It should moved and appended to the Scintilla-message paragraph, because it only applies to types 0 and 1, not to 3 (or all) as might be inferred from the placement.

pryrt:
* merge the "message" paragraphs with the "scintilla message" paragraph, per #149
* clarify that macros are not always sufficient for automation tasks, and recommend alternatives

closes #149


sasumner (approved):
Nice.

dail8859 (commented):
👍

PR #152, opened 2021-01-18
Add info about removing unwanted items from Find combobox histories

### LINKED ISSUE #151: Add instructions about new removing item from find history feature in Find dialog
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9396

sasumner:
Fix #151

I thought about adding a parenthetical "requires 7.9.3 or later" at the end of the last sentence.
If our "documentation manager" thinks that is necessary, I'll do it, although I tend to not like things like that in the docs (not sure why).

Also, added some text about using Alt+down to "drop" a combobox via keyboard as there was a recent user inquiry about it:  https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9404

Regarding this existing verbage:
**_...be careful though, sometimes when editing a control with the history feature, a user will accidentally hit an up or down arrow key when they really mean to press left or right arrow; this unfortunately wipes out the search/replace expression (as those are edited most often) that was being worked on and replaces it with some different text from the history buffer._**
How about a new N++ issue for a new Preference setting which would enable/disable down/up arrow from moving through the history in this manner?  If disabled, user would be required to "drop the box" in order to select an item from the history.


pryrt (approved):
The doc changes look good to me.

pryrt:
> "requires 7.9.3 or later"

I tend to add those for features, because users will be sent to the same manual, whether they are on the most recent Notepad++ or whether they downloaded v7.7 and never updated since.  I understand that it can "clutter" up the docs, but since we don't have an archive of "this version of the manual goes with this version of the program" (and since the manual doesn't release simultaneously with the software), I think the version-note can help.  But I don't require it.

pryrt:
> How about a new N++ issue for a new Preference setting ...

I wouldn't fight against such an issue, but personally I don't think that the benefit would be worth the effort, and would rather see development time go into something more interesting than a preference changing how arrows work in a single dialog.

pryrt:
> The doc changes look good to me.

In case that review-comment wasn't clear, @sasumner can feel free to merge the change in at this point.

sasumner:
> rather see development time go into something more interesting than a preference changing how arrows work in a single dialog.

Yes, me too.
However...recently I was typing a longish regex into the _Find what_ box and I inadvertently hit either up or down (can't remember exactly which).  :-(
So, sometimes what you want to spend time on is given weight by the errors you yourself make.  :-)

sasumner:
>I think the version-note can help

I will add it in then.

Perhaps a policy of periodically going through the docs and removing really old (e.g. "one full cycle" behind) version notes would be a good thing?  ("one full cycle" defined by example: at 7.9.3, all such version notes that are 7.8.3 and older would be removed).  Just a suggestion.

PR #159, opened 2021-02-08
Clarify Column Editor Initial Number

vermiceli:
Add clarity on how the initial number is interpreted.  For example, if
format is hex the initial number of 17 will be outputted as 11 and
not 17.


vermiceli:
I just now saw that this is clarified in the format section, which I missed when reading this document.

PR #170, opened 2021-03-12
CI github action to build the website

chcg:
To store artifacts something like:

```
    - name: Archive artifacts for hugo
      uses: actions/upload-artifact@v1
      with:
          name: N++ page
          path: to_be_defined
```

could be added.

### DIFF
--- a/.github/workflows/CI_build.yml
+++ b/.github/workflows/CI_build.yml
@@ -11,12 +11,20 @@ jobs:
     - name: Checkout repo
       uses: actions/checkout@v2
 
-    - name: Install hugo extended via chocolatey
+    - name: Install hugo extended and httrack via chocolatey
       working-directory: .
-      run: choco install hugo-extended
+      run: choco install hugo-extended httrack
 
-    - name: Build docu
+    - name: Retrieve data from server
       working-directory: .
-      run: hugo.exe --minify --theme book
+      run: |
+           cmd /c start hugo.exe server --minify --theme book
+           $env:Path += ";C:\Program Files\WinHTTrack\" 
+           httrack.exe "http://127.0.0.1:1313/" -O "./nppUserManual" "+*.npp-user-manual.org/*" -v
 
+    - name: Archive artifacts for hugo
+      uses: actions/upload-artifact@v1
+      with:
+          name: N++ static page
+          path: ./nppUserManual/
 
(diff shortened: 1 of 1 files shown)


chcg:
extended to use httrack to get local version, see https://github.com/chcg/npp-usermanual/actions/runs/648074025#artifacts maybe with one dir level 127.0.0.1_1313 to much

pryrt:
Cool.  I did some experimenting in a [dummy repo](/pryrt/action_asset).  
1. I navigated into the folder, and did a rename and made my own zipfile -- benefits:
    1. no httrack-based url wrapper in what we distribute to end users
    2. renames the folder, so the zipfile won't contain 127.0.0.1_1313; instead, the folder will be nppUserManual to match the zipfile name
2. artifact vs asset: 
    * artifacts are good for individual builds, but from the way I understand it, assets are easier to access from tagged releases.  As an asset on a tag, I can provide a direct link to the zipfile with https://github.com/pryrt/action_asset/releases/latest/download/nppUserManual.zip -- and this link will always be the same for the latest tagged release of the website, without having to change the URL in the manual (or know beforehand what the run number is for finding the right artifact)
    * also, by limiting the github action to tags, instead of all push and pull requests, we won't create the zipfiles every time we make an edit on one of our forks/branches, or every time we make a pull request back into the main repository; with this limitation, it will only be when @donho creates a new tag and releases a new version to the live website.

The examples I linked above also added a "download" button to the first page in the website.

@chcg, if you and @donho like the direction my experiments are going, do you want me to edit your PR, or do we want to reject your PR and have me submit a separate one?

chcg:
@pryrt What you did is more like CD (continuous delivery) than CI. The question is if you need the option to check the build results by an independent build instance before creating a tag/release or check the hugo output visually of a PR. For SW builds it is easier to know exactly which checkin was breaking the build but probably that is not the case here.  As I'm not knowing how your workflow well enough, maybe that is not necessary and then a trigger just on tags is sufficient.
For me it is fine to close this PR and created a new one fitting to your needs.

pryrt:
@chcg,  

You have a good point.  We probably need both.  We can use yours for checking the zipfile on any push/pull, and I'll create a separate action that will create the final asset on a tag.

pryrt (approved):
per my comments, let's do both

PR #171, opened 2021-03-14
Add information about "Find in Projects" tab of Find dialog.

Uhf7:
Added at least a basic hint of the existence of the new function. And I tried to make the section introducing dialog based searching more readable, I had to extend it anyway.


pryrt (approved):
Looks good.  I will merge

PR #174, opened 2021-03-14
Create onTag_zip_manual_as_asset.yml

pryrt:
for every tag release: converts the Hugo site to a standalone HTML, zips into nppUserManual.zip, and saves the zip as an asset in the tag


chcg (approved):
Looks good to me. You could use the same build steps also for the CI build.

PR #175, opened 2021-03-16
Github CI action update

chcg:
- added dependabot.yml action to check for updated github actions on a weekly base
- adapted CI_build.yml to use similar build steps as one used for tag action


pryrt (approved):
Looks good.  I will merge

PR #177, opened 2021-03-16
Add link to nppUserManual.zip asset from main page

pryrt:
Apparently, my link wasn't in the branch that I did my last PR on, so I am adding it here.

When I was trying out my original fancy formatting for the link that I used during experiments, it would render in hugo's live server, but not in the httrack extraction.  So I settled on a simple markdown link, more in keeping with the rest of the document.  This link makes it all the way into the httrack extraction in the artifact (thanks @chcg for recommending we keep the CI element of the actions -- it was obviously the right choice), and everything looks good to me.

### DIFF
--- a/.github/workflows/CI_build.yml
+++ b/.github/workflows/CI_build.yml
@@ -11,12 +11,20 @@ jobs:
     - name: Checkout repo
       uses: actions/checkout@v2
 
-    - name: Install hugo extended via chocolatey
+    - name: Install hugo extended and httrack via chocolatey
       working-directory: .
-      run: choco install hugo-extended
+      run: choco install hugo-extended httrack
 
-    - name: Build docu
+    - name: Retrieve data from server
       working-directory: .
-      run: hugo.exe --minify --theme book
+      run: |
+           cmd /c start hugo.exe server --minify --theme book
+           $env:Path += ";C:\Program Files\WinHTTrack\" 
+           httrack.exe "http://127.0.0.1:1313/" -O "./nppUserManual" "+*.npp-user-manual.org/*" -v
 
+    - name: Archive artifacts for hugo
+      uses: actions/upload-artifact@v1
+      with:
+          name: N++ static page
+          path: ./nppUserManual/
 
--- a/content/_index.md
+++ b/content/_index.md
@@ -12,16 +12,6 @@ Of course, it's not our goal to create bad documentation, but you get what we me
 
 **Notepad++ User Manual** is built collaboratively, and [your contribution is very welcome](https://github.com/notepad-plus-plus/npp-usermanual).
 
-<center>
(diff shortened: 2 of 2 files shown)


chcg (approved):

donho:
@pryrt 
How will we generate the zip file?

sasumner:
TBH I'm not following this discussion that closely, but it is making me wonder if there is value in distributing the manual (in such a form) at N++ release time, along with the portable release at least?

donho:
@sasumner 
> but it is making me wonder if there is value in distributing the manual (in such a form) at N++ release time, along with the portable release at least?

It's always good to distribute the manual with the package, though I think the most of users won't read it. My concern is the distributed package size.  Also that will make release procedure more painful. So let's keep it as it as - if user needs to RTFM, they can go online , otherwise they can download the zip format offline manual.

pryrt:
> How will we generate the zip file?

@donho , when you create a new npp-usermanual tag, the GitHub action will trigger; it will run hugo and httrack, create the zipfile, and add it as an "asset" to the tag that you are creating -- all without you having to do anything (I wanted it to be automatic).  

The zipfile for the most recent usermanual release (tag) will always be available at the URL https://github.com/notepad-plus-plus/npp-usermanual/releases/latest/download/nppUserManual.zip , so it's a nice consistent URL.  (Where "always" is defined as after the first tag which includes this GitHub action, so as of the next tag you create.  If you want, I could manually upload a zipfile for the currently most-recent tag at https://github.com/notepad-plus-plus/npp-usermanual/releases/tag/2.3 )

--

For debug purposes, every time a push or pull is done to a npp-usermanual fork, branch, or main repo -- including all pull requests to the main repo -- the same process is automatically used to create an artifact for the given build run; so you can go to the "checks" page (for example, https://github.com/notepad-plus-plus/npp-usermanual/pull/177/checks), and go to the Artifacts and download the zipfile generated from that build.  
image
With this artifact, we can make sure that httrack+hugo agrees with what we saw when developing our edits on a local hugo server.

pryrt:
@sasumner said,
> making me wonder if there is value in distributing the manual (in such a form) at N++ release time

As with @donho, I think that would complicate the release procedure.  Besides, the usermanual updates often lag at least the RC, and often even the full release, by a week or more (depending on how many changes are involved).

I think having the usermanual zipfile automatically created with each npp-user-manual.org release is a good compromise.  

The download link for the zipfile will be available on the first page of the npp-user-manual.org website; if that's not convenient enough, you might convince Don to also add a new **?**-menu entry of **Notepad++ Offline Documentation Download** to go next to **Notepad++ Online Documentation**

sasumner:
No alterations in the process needed, I just thought I'd float it as an idea.

PR #179, opened 2021-03-20
Fix the description of "Replace All"

### LINKED ISSUE #178: Inaccuracy in "Replace All"
See discussion in [forum: "New Line every n characters"](https://community.notepad-plus-plus.org/topic/20916/)

Basically, manual says "Replace All: Replaces all matches from the active location onward (following the Backward direction and Wrap around settings as appropriate)."  But it appears that "wrap around" being checked will change the behavior subtley.

```
one two three four five
fourteen something fifteen
```
* FIND = `(?i-s)\b((?<!☺ )f\w+)`
* REPLACE = `${1}☺`

(so it will only replace words starting with "f" that don't have smiley-space before it)

Put the cursor between `four` and `five`.  

If wrap-around is disabled, then **Replace All** will result in
```
one two three four five☺
fourteen☺ something fifteen☺
```
... which means the search really did start between the `four` and the `five`.   This is the behavior I expected.

But with wrap-around enabled, then **Replace All** will result in
```
one two three four☺ five
fourteen☺ something fifteen☺
```
... which says that the search must have started at the beginning of the file, rather than starting between `four` and `five`.  This surprised me, and seems to contradict the phrasing in the manual.

@sasumner , can you (or someone else familar with the codebase) confirm that wrap-around really does cause it to start from the beginning of the file for the global replacement?

pryrt:
Include the lessons reiterated / relearned in https://community.notepad-plus-plus.org/topic/20916/new-line-every-n-characters,
as well as Guy's table from https://community.notepad-plus-plus.org/post/27946

closes #178

### DIFF
--- a/content/docs/searching.md
+++ b/content/docs/searching.md
@@ -70,7 +70,7 @@ The various action buttons available include:
 * **Close**: Closes the search dialog
 
 * **Replace**: Replaces the currently-selected match.  (If no match is currently selected, it behaves like **Find Next** and just highlights the next match in the specified direction)
-* **Replace All**: With **☑ Wrap Around** enabled, it makes one pass through the active document, from the very top to the very bottom, and replaces all occurrences found.  With **☐ Wrap Around** disabled, it searches from the caret to the end of the file (if **☐ Backward direction** is disabled) or from the caret to the beginning of the file (if **☑ Backward direction** is enabled) and replaces all occurrences found in that region.
+* **Replace All**: With **☑ Wrap Around** ticked, it makes one pass through the active document, from the very top to the very bottom, and replaces all occurrences found.  With **☐ Wrap Around** unticked, it searches from the caret to the end of the file (if **☐ Backward direction** is unticked) or from the beginning of the file to the caret (if **☑ Backward direction** is ticked) and replaces all occurrences found in that region.
     * NOTE: for regular expressions, this will be equivalent to running the regular expression multiple times, which is _not_ the same as running with the /g global flag enabled that is available in the regular expression engines of some programming-languages.
     * To clarify the **Replace All** results, depending on the condition of the various settings:
 
--- a/content/docs/searching.md
+++ b/content/docs/searching.md
@@ -70,7 +70,7 @@ The various action buttons available include:
 * **Close**: Closes the search dialog
 
 * **Replace**: Replaces the currently-selected match.  (If no match is currently selected, it behaves like **Find Next** and just highlights the next match in the specified direction)
-* **Replace All**: Replaces all matches in the active document.  With **☑ Wrap Around** enabled, it makes one pass through the active document, from the very top to the very bottom.  With **☐ Wrap Around** disabled, it searches from the caret to the end of the file (if **☐ Backward direction** is disabled) or from the caret to the beginning of the file (if **☑ Backward direction** is enabled).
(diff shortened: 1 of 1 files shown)


content/docs/searching.md
```diff
@@ -70,8 +70,21 @@ The various action buttons available include:
 * **Close**: Closes the search dialog
 
 * **Replace**: Replaces the currently-selected match.  (If no match is currently selected, it behaves like **Find Next** and just highlights the next match in the specified direction)
-* **Replace All**: Replaces all matches from the active location onward (following the **Backward direction** and **Wrap around** settings as appropriate).
+* **Replace All**: Replaces all matches in the active document.  With **☑ Wrap Around** enabled, it makes one pass through the active document, from the very top to the very bottom.  With **☐ Wrap Around** disabled, it searches from the caret to the end of the file (if **☐ Backward direction** is disabled) or from the caret to the beginning of the file (if **☑ Backward direction** is enabled).
```
sasumner, line 73: It leads off with "Replaces all matches in the active document".  But then goes on to explain how that may not be true.

pryrt: Removed that sentence and tweaked the next two sentences to be more accurate.

content/docs/searching.md
```diff
@@ -70,7 +70,7 @@ The various action buttons available include:
 * **Close**: Closes the search dialog
 
 * **Replace**: Replaces the currently-selected match.  (If no match is currently selected, it behaves like **Find Next** and just highlights the next match in the specified direction)
-* **Replace All**: Replaces all matches in the active document.  With **☑ Wrap Around** enabled, it makes one pass through the active document, from the very top to the very bottom.  With **☐ Wrap Around** disabled, it searches from the caret to the end of the file (if **☐ Backward direction** is disabled) or from the caret to the beginning of the file (if **☑ Backward direction** is enabled).
+* **Replace All**: With **☑ Wrap Around** enabled, it makes one pass through the active document, from the very top to the very bottom, and replaces all occurrences found.  With **☐ Wrap Around** disabled, it searches from the caret to the end of the file (if **☐ Backward direction** is disabled) or from the caret to the beginning of the file (if **☑ Backward direction** is enabled) and replaces all occurrences found in that region.
```
sasumner, line 73: Personally I don't like the "disabled"/"enabled" verbiage.  Is "unasserted"/"asserted" better?  Or "ticked"/"unticked"...  

I would change the backward-dir part to be "from the beginning of the file to the caret" since that is what is actually occurring.

pryrt: rephrased again...

sasumner (approved):

sasumner:
So _Mark_ uses the _Wrap around_ checkbox in a similar way to _Replace All_.
That is, to mean "whole document".
Do you think the documentation describing _Mark_ is adequate?
(These situations are awkward -- meaning that I haven't checked -- but I could certainly actually go and check instead of posing the question)

pryrt:
> So _Mark_ uses the _Wrap around_ checkbox in a similar way to _Replace All_.
> That is, to mean "whole document".
> Do you think the documentation describing _Mark_ is adequate?
> (These situations are awkward -- meaning that I haven't checked -- but I could certainly actually go and check instead of posing the question)

_Mark All_ had almost no description.  I updated it to match the description of _Replace All_

sasumner (approved):

guy038:
Hello, @sasumner and @pryrt,

Here is an **updated** version of my table, as the **`Count`** and **`Mark`** options have the **same** behaviour than the  **`Replace All`** option !

https://community.notepad-plus-plus.org/post/64236

BR

guy038

PR #181, opened 2021-03-22
rework the contextMenu section

### LINKED ISSUE #180: contextMenu: need slightly more info
prompted by https://community.notepad-plus-plus.org/topic/20928/

The one sentence on "If you wish to use IDs" doesn't have quite enough information to make it useful.  

Need to describe `<Item id="0"/>` being used as separator

The main bullets should have three points, not just two: builtin commands by name, builtin commands by id, and plugin commands.

pryrt:
It is now more explicit that by-name and by-ID are separate ways of calling the same menu entry.  It gives examples (derived from the default contextMenu.xml).

Also, it clarifies what happens if there is no `ItemNameAs` or `FolderName` attribute.

closes #180

### DIFF
--- a/content/docs/config-files.md
+++ b/content/docs/config-files.md
@@ -48,9 +48,9 @@ The context menu does not have a GUI-based editor; you just need to edit the fil
 All menu commands can be added to the Context Menu, including plugin commands:
 
 * To add a built-in command, you may provide the main menu name (as it appears in the main menu bar) as the value of the MenuEntryName attribute and the command's item name (as it appears in the menu) as the value of the MenuItemName attribute. The MenuEntryName attribute must reference an entry on the main menu bar and must be an ancestor of the MenuItemName attribute, regardless of its depth.  
-    * For example, `<Item MenuEntryName="Edit" MenuItemName="Cut"/>` will add a context-menu entry that calls the **File** menu's **Cut** entry.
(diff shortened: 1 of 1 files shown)


content/docs/config-files.md
```diff
@@ -47,18 +47,25 @@ The context menu does not have a GUI-based editor; you just need to edit the fil
...
-* To add a plugin command, you need to provide the plugin's menu item name (as it appears in the Plugins menu) as the value of the PluginEntryName attribute and the command's menu item name (as it appears in the plugin's sub-menu) as the value of the PluginCommandItemName attribute.
+* To add a built-in command, you may provide the main menu name (as it appears in the main menu bar) as the value of the MenuEntryName attribute and the command's item name (as it appears in the menu) as the value of the MenuItemName attribute. The MenuEntryName attribute must reference an entry on the main menu bar and must be an ancestor of the MenuItemName attribute, regardless of its depth.  
+    * For example, `<Item MenuEntryName="Edit" MenuItemName="Cut"/>` will add a context-menu entry that calls the **File** menu's **Cut** entry.
```
sasumner, line 51: should be "the **Edit** menu's **Cut** entry"

pryrt: whoops.

pryrt: @sasumner , other than the whoops (now fixed), is it okay?

sasumner (approved):

Issue #176, opened 2021-03-16
Convert "Search results Window" into separate chapter


Uhf7:
I have a little problem here with the structure of the manual. It regards the chapter "searching".

As it is now:

Searching
- Dialog based searching
  - Find / Replace
  - Find in Files
    with a large sub-section "Search results" window
  - Find in Projects
  - Highlighting and Bookmarking
- Dialog-free search/mark actions
  - Searching
  - Highlighting
- Finding characters in a specific range
- Incremental Search  ...
   ...

Within the sub-chapter "Find in Files" there is embedded a long section about the ""Search results" window".  This section is important, and it actually applies now to two higher-ranking sections, "Find in Files" and "Find in Projects". 

So my proposal is here to move the section ""Search results" window" into an own chapter, at the same level with "Dialog-based Searching", "Dialog-free search/mark actions", and "Finding characters in a specific range".

As it should be:

Searching
- Dialog based searching
  - Find / Replace
  - Find in Files
  - Find in Projects
  - Highlighting and Bookmarking
- "Search results" window
- Dialog-free search/mark actions
  - Searching
  - Highlighting
- Finding characters in a specific range
- Incremental Search  ...
   ...

Please tell me, if there is anyone out there who feels the same way ...

pryrt:
Given that the results window can apply to **Find/Replace**, **Find in Files**, and now **Find in Projects**, it makes sense to me to move it out from under **Find/Replace**.  

But, since the **"Search Results" Window** is a result of the Dialog-based searching, maybe it should be a fifth subsection section, rather than an independent main section:
* Dialog based searching
    * Find / Replace
    * Find in Files
    * Find in Projects
    * Highlighting and Bookmarking
    * "Search results" window

... but I'm not heavily invested either way, and I understand the counter-argument that "Search results" is not one of the tabs on the main dialog, whereas the other four sub-sections are.  

So I'd be fine with either making it the fifth subsection (inside **Dialog based searching**), or making it the second main section (after **Dialog based searching**).

@sasumner was heavily involved in the GUI portions of the searching.md page, so getting his opinion would be good.

sasumner:
@pryrt 

Thanks for asking my opinion.

I agree that it deserves its own section, either as a "fifth" point of dialog-based searching, or as more of a "main" searching section.  It makes sense either way; perhaps I have a slight leaning towards not making it a 5th point.

Uhf7:
[...]

PR #182, opened 2021-03-26
Move section "Search results window" one level up.

Uhf7:
Implements #176 
- Moved section "Search results window" one level up, so that it appears as separate chapter under "Dialog-based Searching".
- Renamed the other chapters in "Dialog-based Searching" as suggested here: https://github.com/notepad-plus-plus/npp-usermanual/issues/176#issuecomment-801117415


pryrt (commented):
It looks good to me, but Scott should review, too.

sasumner (approved):
Looks like a worthwhile reorg to me.

PR #187, opened 2021-03-30
document -titleAdd command line parameter

### LINKED ISSUE #185: Describe -titleAdd command line argument
See https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9704 "Add command line parameter to add string to title bar"

sasumner:
Fix #185

Hmmm, didn't do it quite the same way as one will see if they do `?` > `Command Line Arguments...` in N++ itself.
But I suppose it is acceptable.
I think it is better described the way I did it in the manual docs (and isn't that the way it should be -- what would be the point of just repeating verbatim?)


pryrt (approved):
go ahead and merge

PR #195, opened 2021-04-14
"Nimrod" -> "Nim" (language was renamed 2014)

trustable-code:


sasumner:
So this would rename it in the documentation, but what about in the actual program; some occurrences there:

image

image

image

trustable-code:
Ok, this should be changed too, unfortunately I have no idea how this works in the Notepad++ source code.

sasumner:
>Ok, this should be changed too, unfortunately I have no idea how this works in the Notepad++ source code.

Can you open an issue regarding it in the N++ code site?  You can link over here, tying the 2 things together.

pryrt (approved):
Looks good, since notepad-plus-plus/notepad-plus-plus#9768 was fixed, I will merge this one, too

PR #197, opened 2021-04-16
Update searching.md

Ekopalypse:
Start and end buffer anchors, are swapped.

### DIFF
--- a/content/docs/searching.md
+++ b/content/docs/searching.md
@@ -525,7 +525,7 @@ Anchors match a zero-length position in the line, rather than a particular chara
 
 * `\A` or `` \` `` ⇒ Matches the start of the file.
 
-* `\z` or `\'` <!-- ` --> ⇒ Matches the end of the file.
+* `\z` or `\'` ⇒ Matches the end of the file.
 
 * `\Z` ⇒ Matches like `\z` with an optional sequence of newlines before it. This is equivalent to `(?=\v*\z)`, which departs from the traditional Perl meaning for this escape.
 
--- a/content/docs/searching.md
+++ b/content/docs/searching.md
@@ -523,9 +523,9 @@ Anchors match a zero-length position in the line, rather than a particular chara
 
 * `\B` ⇒ Not a word boundary. It represents any location between two word characters or between two non-word characters.
 
-* `\A` or `\'` ⇒ The start of the matching string.
+* `\A` or `` \` `` ⇒ The start of the matching string.
 
-* `\z` or `` \` `` <!-- ` --> ⇒ The end of the matching string.
+* `\z` or `\'` <!-- ` --> ⇒ The end of the matching string.
 
 * `\Z` ⇒ Matches like `\z` with an optional sequence of newlines before it. This is equivalent to `(?=\v*\z)`, which departs from the traditional Perl meaning for this escape.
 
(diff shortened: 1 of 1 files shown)


Ekopalypse:
Fixes  #198

sasumner:
Reference is here: https://www.boost.org/doc/libs/1_75_0/libs/regex/doc/html/boost_regex/syntax/perl_syntax.html

Scroll down to see:

image

content/docs/searching.md
```diff
@@ -523,9 +523,9 @@ Anchors match a zero-length position in the line, rather than a particular chara
 
 * `\B` ⇒ Not a word boundary. It represents any location between two word characters or between two non-word characters.
 
-* `\A` or `\'` ⇒ The start of the matching string.
+* `\A` or `` \` `` ⇒ The start of the matching string.
```
pryrt, line 526: Could you update per [Alan's comments](https://community.notepad-plus-plus.org/post/65053): "The start of the file"

content/docs/searching.md
```diff
@@ -523,9 +523,9 @@ Anchors match a zero-length position in the line, rather than a particular chara
...
 
-* `\z` or `` \` `` <!-- ` --> ⇒ The end of the matching string.
+* `\z` or `\'` <!-- ` --> ⇒ The end of the matching string.
```
pryrt, line 528: ... and "The end of the file."

pryrt (changes requested):

pryrt:
I suggested using "file" instead of "matching string", because "matching string" was just wrong (or at least inherently misleading).  And while the boost docs quoted use "buffer", I don't think that most Notepad++ users would think in terms of "buffers".  Since it's effectively the file during Notepad++ usage, I think that's how it should be phrased.

content/docs/searching.md
```diff
@@ -523,9 +523,9 @@ Anchors match a zero-length position in the line, rather than a particular chara
 
 * `\B` ⇒ Not a word boundary. It represents any location between two word characters or between two non-word characters.
 
-* `\A` or `\'` ⇒ The start of the matching string.
+* `\A` or `` \` `` ⇒ Matches the start of the file.
 
-* `\z` or `` \` `` <!-- ` --> ⇒ The end of the matching string.
+* `\z` or `\'` <!-- ` --> ⇒ Matches the end of the file.
```
sasumner, line 528: What's with the `<!-- ... -->` part of the `\z` line?
I realize that it is not a change here, just wondering its purpose.

Ekopalypse: Unnecessary comment - cleaned up.

pryrt: As far as I know, markdown accepts HTML comments (just like it technically accepts other HTML syntax).  (Even Notepad++'s "Markdown (preinstalled)" recognizes HTML comments as comments.

However, those comments were there because apparently my Markdown UDL has a bug/deficiency in it, so it would extend the backtick-highlighting across many lines on the line with the escaped backtick inside the backticks.  I see that the "Markdown (preinstalled)" doesn't have that problem.  Interesting.  So yes, they were something I had just to make the formatting look right in Notepad++ while I was editing.  I guess I'm going to have to switch over to the builtin Markdown UDL rather than my own.

pryrt (approved):

PR #214, opened 2021-05-10
clarify buffer allocation in lexer plugin part

Ekopalypse:


content/docs/plugins.md
```diff
@@ -117,7 +117,7 @@ Description:
 	This function should return the name which should be displayed in the LanguageMenu.   
 	If more than one lexer is contained in a plugin, this function will be called accordingly often  
 	and the respective lexer is queried via index.   
-	Name is a pointer to the memory area allocated by Npp.   
+	Name is a pointer to the memory area allocated by Npp. 8bit encoded name is expected.
```
pryrt, line 120: The phrase "8bit encoded name is expected" is awkward.
I might suggest "This name should be encoded as an 8bit string using the current Windows codepage."

content/docs/plugins.md
```diff
@@ -132,7 +132,8 @@ Args:
 Return: none  
 Description:
 	This function should return the description to be displayed in the first field of the status line.  
-	The parameters are to be treated in the same way as in the GetLexerName function.  
+	The parameters are to be treated in the same way as in the GetLexerName function, except for 
+	the name parameter, here a utf16 encoded name is expected.
```
pryrt, line 136: maybe change "here a utf16 encoded name is expected" to "which requires a string encoded with UTF-16."

pryrt (changes requested):
suggested slight tweaks to phrasing...

Ekopalypse:
@pryrt updated, thanks for the review

pryrt (approved):
Thanks!  I will merge it.

Issue #216, opened 2021-05-12
Document reverse line order feature


sasumner:
References:

* https://github.com/notepad-plus-plus/notepad-plus-plus/commit/ddd4448192f8dcbc72accb20a0a077dbdfc27c21
* https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9839 "Add ability to reverse line order"
* https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9837 "[Feature request] Reverse line order"

debiedowner:
Will do, once the feature makes it into release.

Or maybe I can add it earlier? But I figured adding it might be confusing if next version takes long to release; also it seems I need to mention the version it was added, I don't know if I can assume it will be called 7.9.6. It looks like some dark theme stuff is coming, so maybe it will have a more major release name instead?

I probably should have gone ahead and added it anyway instead of writing this comment as it would be much shorter :) Anyway, I'll remember to do it.

sasumner:
@debiedowner 

>Or maybe I can add it earlier?

Go for it! :-)
You do not have to wait for it to "make it into release".
The user-manual release person won't release a new manual until the appropriate time.

We typically add some words referring to the release it is in, in the manual entry, example:
"feature xxx is yyy (available in v7.9.6 and later)"

>if I can assume it will be called 7.9.6

If this changes the manual release person will alter it later.
You can assume 7.9.6.

debiedowner:
Oh, I mistakenly thought that the manual was just a Github Page for this repo, so anything merged would go live immediately; that's why I was wary of making a change prematurely. I now realize that it is separately generated using Hugo. I am making a pull request now.

PR #220, opened 2021-05-15
Document reverse line order feature

debiedowner:
Implements #216


sasumner:
@pryrt Please let me review this one.  I'm assigning it to myself.

sasumner:
@debiedowner 

I had some offline discussion with the Notepad++ author about this feature and we decided to make this change of location for the menu entry:

image

We decided to move the "random sort" as well, and change its text slightly, so the sum total of all changes will appear like this:

image

Because this was my suggestion, and my code to change it (in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9868 ), I'll take over finishing the documentation for all of this.

So I'm going to close your PR without acceptance.
Sorry for the wasted time you went through to create it.
But...and this is important:  Thank you for providing user manual input for your code change.

debiedowner:
@sasumner Makes sense, neither are actually sorts; I only placed it there to follow the precedent of "Sort Lines Randomly". I initially even considered naming it "Sort Lines Reversely" to fit in with the others, but it makes even less sense than the 'randomly' one, as it is even less of a sort, being dependent on the existing line order. So instead I considered renaming the 'randomly' item exactly like you suggest, though I didn't, to keep my pull request as minimal as possible.

sasumner:
@debiedowner 

Yep, sometimes an addition makes something else not "fit right".  I think that's the case here.  But...we then make an adjustment to make the evolution better.

PR #239, opened 2021-06-04
Fix link

steve02081504:


steve02081504 (commented):
(⊙o⊙)？

pryrt (approved):
I would've sworn we'd fixed that link some time in the past.

Ah, well. Thanks for catching that.

PR #242, opened 2021-06-04
Document Advanced Setting nbMaxFindHistoryFind

### LINKED ISSUE #241: Document Advanced Setting nbMaxFindHistoryFind
Per https://community.notepad-plus-plus.org/post/66702, and historically https://github.com/notepad-plus-plus/notepad-plus-plus/issues/2944#issuecomment-384406464 and https://github.com/notepad-plus-plus/notepad-plus-plus/issues/4511, all indicate that users occasionally want this feature, but aren't aware of how to set `nbMaxFindHistoryFind` and the related `nbMaxFindHistoryReplace` and `nbMaxFindHistoryPath` and `nbMaxFindHistoryFilter`.  Since that setting has been around (but hidden) since at least 2007, we should probably document it. :-)

pryrt:
closes #241 

* describes that (and related) settings in Preferences for Advanced Users
* in Searching, link from the mention of "number of search/replace terms retained may be changed by editing the config.xml configuration file" to the Preferences for Advanced Users
* in config-files Editing Configuration Files section, clarify that changes to `config.xml` require a more drastic workaround than the other configuration files (confirmed while verifying whether or not "0" would work above.


content/docs/preferences.md
```diff
@@ -517,6 +517,12 @@ The following settings are for rather specific needs and could cause some confus
 ```
 Simply add `"regexBackward4PowerUser"="yes"` if this option is absent.
 
+* Change the number of expressions that can be saved in the [Find/Replace](../searching/) dialog's Find, Replace, Filters, and Directory fields.  They are set in the `<FindHistory ...>` tag inside `config.xml`, using the attributes:
+    * `nbMaxFindHistoryFind` => changes the number of **Find what** entries that are saved in the history (0 .. 30, default = 10)
+    * `nbMaxFindHistoryReplace` => changes the number of **Replace with** entries that are saved in the history (0 .. 30, default = 10)
+    * `nbMaxFindHistoryPath` => changes the number of **Filters** entries that are saved in the history (0 .. 30, default = 10)
+    * `nbMaxFindHistoryFilter` => changes the number of **Directory** entries that are saved in the history (0 .. 30, default = 10)
```
sasumner, line 524: Here's where they are defined: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f9d6fb9e31719a0ef9bbc8bd42c262560575adf8/PowerEditor/src/Parameters.h#L982

Note that the default for the filters is 20, not 30 as you have here in the docs.

sasumner (changes requested):
See comment about number of filters

content/docs/preferences.md
pryrt: per https://github.com/notepad-plus-plus/notepad-plus-plus/blob/af56713079c9e31b8c81b5603a6c841afb3e2005/PowerEditor/src/config.4zipPackage.xml#L50 , the default value in the config file is 10 for all four.  The line you point to actually indicates that for the filter, the MAX value is 20.  Unless I've read things wrong.

sasumner: My brain is gone:  I meant to say "max" not "default".
The max for the filters is the only one of the four that is not 30; it is 20; that's what my source code reference was about.
Your preferences.md change has 30 for all four.  That was the complaint.

sasumner (changes requested):
I think this one is on you? :)
You changed "path" to 20; should have been "filter".

pryrt:
It's because path and filter were swapped in name vs description.  I'll fix that.

sasumner (approved):

PR #245, opened 2021-06-06
Tweak the Editing Config Files

pryrt:
I don't like that #242 recommended using notepad.exe to edit Notepad++'s `config.xml`, so I re-phrased it to supply that as an alternative, but clarified that using a portable copy of Notepad++ would work as well, and would avoid having to use some other brand of editor, even for editing `config.xml`. :-)


sasumner (approved):
I think it "muddies the waters" and makes it a TLDR.  Nobody but us "pros" are going to have a "portable" on hand for such things.
Of course, only a "pro" would consider tweaking the config files anyway.
All right, it's OK.

PR #252, opened 2021-06-26
Clarify periodic-backup file-naming conventions

### LINKED ISSUE #250: Periodic Backup file naming
In this [forum post](https://community.notepad-plus-plus.org/post/67302), Alan liked the way that I had described the periodic backup feature, especially with regards to file naming, and asked me to consider adding some of that information into the user manual.

pryrt:
closes #250


sasumner (changes requested):
Nice job.
How about a change to the Backup-on-save Verbose section, with the same "feel" as this one?
Add a new first bullet point ahead of the existing one:  "if a Custom Backup Directory is defined,..." and proceed to give a nice example with a 2021-06-23... file.
I think you get what I mean; sorry for upscoping your change here; I'd do it after you're done, but, well, you've got it "open". :-)

pryrt:
Good idea.
While implementing, I realized the Custom Backup Directory can cause naming collisions on the backup files, so emphasized that in both Simple and Verbose as well.

sasumner (approved):

PR #267, opened 2021-07-23
clarify "remember last used directory" setting

pryrt:
@mere-human, is this sufficient?


mere-human:
> @mere-human, is this sufficient?

@pryrt Yes, this looks much better now!

mere-human (approved):

PR #271, opened 2021-08-04
Update auto-completion.md

jonmpayne:
Added note for creating custom suggestions for the default language, Normal Text.


pryrt (approved):
thanks

PR #277, opened 2021-08-21
Properly format the Readme and contributing guidelines using prettier-vscode.

sanikdah:
Also fixes that fact that the codeblock at the end of the README was being interpreted as bash CODE.  Causing weird coloring that I don't think is wanted.  Feel free to let me know if you want that added back for some reason.


pryrt:
@sanikdah ,

I cannot figure out a reason for the change you made in this PR.  It worked as-is.  And you might disagree with the choices of `*` vs `-` for bulleted lists, or `*italic*` vs `_italic_`, but those are all perfectly-valid markdown, and I see no reason to change them just because some random user disagrees with our choice of markdown...

Can you please explain what you are trying to actually accomplish?

pryrt:
Please see the comments in PR #278 for an explanation of why most of your commits were summarily rejected.

pryrt:
Per CONTRIBUTION.md, one of the guidelines is "Respect Notepad++ User Manual style."  Mass changes to the style is definitely _not_ respecting the style.  Sorry.

PR #278, opened 2021-08-21
Auto Format some more files.

sanikdah:
Some more auto formatting, gonna be doing only a few files per PR, the next PR I make will show why.


sanikdah:
Sorry, I didn't mean to submit that last one to here, let me fix that quick.

pryrt:
@sanikdah ,

This PR, and all your subsequent ones, all contain all the same changes from the first PR, along with whatever else you wanted to modify in the particular PR.  It makes it impossible to tell exactly what you were intending with this or subsequent PRs.  As such, if we were to approve the first PR (which is not guaranteed), then all of these PRs would have conflicts; and if we _didn't_ approve the first PR, all of these would have to be rejected because they contain changes that were rejected.

Further, as in the first PR, I cannot tell what you are hoping to accomplish with any of these PR, other than imposing your own markdown preferences on us.

In general, as a first-time or non-regular contributor to an open source project, it's good to start with an _issue_ rather than a pull request, to make sure you have the buy-in of the team _before_ you spend the effort on the PR; it's also important to keep changes limited in scope, and that any given PR doesn't replicate changes you already made in another PR.  (If you want to make N different changes, you should fork the main repo into your own fork, and then make a branch from the _unmodified_ fork-main, and only change your branches, and make your PR from those separate branches; that will keep the changes properly independent.)

If you cannot give a good justification for what you're doing, and clean up the PRs so they are independent of each other and don't include the changes you made in the others, it is highly likely that all of these requests will be summarily rejected as not meeting the standards of the manual.

pryrt:
Sorry, as I studied your changes more, the vast majority are to the theme which was chosen for the user manual; there is no good reason to change a well-established theme which the owner of the usermanual chose as the preferred theme.  It doesn't matter if you prefer double quotes to single quotes.  I am summarily rejecting all the edits to the theme.  Sorry.

PR #282, opened 2021-08-21
Auto Format all the manual content now too :)

sanikdah:


pryrt:
Per CONTRIBUTION.md, one of the guidelines is "Respect Notepad++ User Manual style."  Mass changes to the style is definitely _not_ respecting the style.  Sorry.

PR #306, opened 2022-01-17
Langs.xml commentLine/commentStart/commentEnd attributes

### LINKED ISSUE #305: confusion over langs.xml commentLine/commentStart/commentEnd attributes
per https://community.notepad-plus-plus.org/topic/22390/wrong-syntax-highlightning, there is still confusion about the commentLine/commentStart/commentEnd attributes for each language in langs.xml.

The usermanual should be updated to clarify that, in both the editing and config-files documents (with cross-linking)

pryrt:
clarifies the role of commentLine/commentStart/commentEnd attributes in langs.xml: 
* explains they are used for the Comment/Uncomment commands in the Edit menu
* explains that changing those in the langs.xml will _not_ affect what syntax the lexers recognize as comments

closes #305


pryrt:
@ArkadiuszMichalski , since you were involved in the [original forum discussion](https://community.notepad-plus-plus.org/topic/22390/wrong-syntax-highlightning), could you give these changes a review, to make sure I have properly interpreted and explained the attributes, and what they do and don't affect?

Thanks.

ArkadiuszMichalski (approved):
@pryrt 
Looks good. Maybe it's also worth adding that these attributes can only define one type of line and block comment. This is relevant for languages that support different variants of comments. The lexer can cover multiple comments, but  insert/ toggle works only for one variant. If the language has many variants, you have to decide for yourself which ones you prefer for insert/ toggle via this three attribiutes.

PR #327, opened 2022-03-09
plugins: add description of the new JSON attribute

pryrt:
per https://github.com/notepad-plus-plus/nppPluginList/pull/423

@rdipardo, please review my changes, and let me know if it needs improvement.

### DIFF
--- a/content/docs/plugins.md
+++ b/content/docs/plugins.md
@@ -333,6 +333,33 @@ and the rest should be intuitive.
    name as the **folder-name** and the plugin DLL file should be placed at the root
    level of the ZIP file. Otherwise Plugin Admin won't install it. Any additionals
    files (DLL or data) can be placed at the root level or in an arbitrary subfolder.
+7. Compatibility: In March 2022, a new parameter was added to the JSON file format, 
+   to indicate your plugin's compatibility with various versions of Notepad++, called
+   `npp-compatible-versions`.  This will have the format 
+   ```
+   "npp-compatible-versions": "[minimum,maximum]",
+   ```
+   where _minimum_ and _maximum_ are version numbers; if _minimum_ is empty, it will
+   mean that the plugin will work with any version of Notepad++ before (and including) the _maximum_;
+   if _maximum_ is empty, it will mean that the plugin will work with any version of
+   Notepad++ after (and including) the _minimum_; you cannot set both empty.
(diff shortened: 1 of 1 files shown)


rdipardo:
@pryrt
Nice work.
I'm afraid, however, it's been obsoleted by notepad-plus-plus/notepad-plus-plus@a84cbe8f

### TL;DR
The `"npp-compatible-versions"` property will not be sufficient to avoid crashes.
Every plugin author must _also_ provide something called `"old-versions-compatibility"`, in the form of `[previous version(s)][previously compatible Npp versions]`, e.g. (for mimeTools), `"[,2.6][,8.2.1]"`

### Background 
There was a chance to make version checking _logical_ and _stable_, as I futilely tried to explain in past days:

- https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11358#issuecomment-1062692927
- https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11356#issuecomment-1061723215

Instead we now have a "system" that depends on plugin authors _telling_ N++ "this one is good, _that one_ is bad."

And when a plugin author forgets the other half, or gets it wrong? I guess we just tell users, "Oh, someone forgot/mistyped the 'old-versions-compatibility' thing"

- And what "someone" is that?
- Dunno; try removing plugins one-by-one to find out!

With "stability" like that, I'd say we're right back to where we started before notepad-plus-plus/nppPluginList#416.

pryrt:
I think I've updated the PR to describe that second parameter as well.  Let me know if it still is incomplete or inaccurate, or whether the needs change again before the next Notepad++/nppPluginsList release.  (I will wait to merge this PR until the next release.)

rdipardo:
- Looks good. (Can professional documentation turn sloppy, impromptu additions to the schema into a "specification"? Time will tell.) 

- I can only vouch for the patterns accepted by the "npp-compatible-versions" regex. I have not and will not be testing the "old-versions-compatibility"` thing. There are [online tools](https://regex101.com) for that.

- Requests for edits are pouring in from plugin authors at notepad-plus-plus/nppPluginList#424. At least wait for the final version to get merged. I predict there will be structural changes made after this hits end user PCs, so _well after_ the next release might be prudent

donho (approved):

donho:
Thank you @pryrt !

PR #346, opened 2022-04-08
Change "Non ASCII" to "Character codepoints"

pryrt:
Clarify that the `\x` and `\0` notations work for any characters, not just non-ASCII.
Add in the codepoint > U+FFFF instructions (thanks to Guy's historical herculean efforts in forum posts)


alankilborn:
I took a look at the changes here.  I even followed the surrogate pair conversion steps to make sure it worked and read well -- it does on both counts.  I didn't see anything worth changing in what was submitted.

Nitpick:  Ok, well maybe there is one too many spaces right before `For example, \0101 looks for the letter A`

PR #351, opened 2022-04-20
v8.4: use ILexer5 for lexer plugins

pryrt:


pryrt:
@Ekopalypse , does this seem like enough about iLexer5 for now?  ... At least until someone has enough experience with iLexer5 to do a similar level of detail to your iLexer4 description in the manual?

Ekopalypse:
@pryrt - GetLexerFactory, GetLibraryPropertyNames, SetLibraryProperty and GetNameSpace are not yet used by Notepad++. 
The only function I had to add to make my VLang lexer work was CreateLexer and return the pointer to the ILexer5 interface.

@DonHo is the current implemented usage how Npp will use it, or are there plans to use/implement the other functions as well? By this I mean mainly the functions GetLibraryPropertyNames, SetLibraryProperty and GetNameSpace. 
GetLexerFactory is obsolete.

donho:
> GetLexerFactory, GetLibraryPropertyNames, SetLibraryProperty and GetNameSpace are not yet used by Notepad++.

I confirm, and they will be optional if they are added in the future.

> is the current implemented usage how Npp will use it, or are there plans to use/implement the other functions as well? By this I mean mainly the functions GetLibraryPropertyNames, SetLibraryProperty and GetNameSpace.

It depends on the need of plugin authors. I will add them if there are the feature requests for these APIs.

> GetLexerFactory is obsolete.

How come?

Ekopalypse:
As far as I know, GetLexerFactory does basically the same as CreateLexer and was introduced to ease the transition from ILexer4 to 5.
Where CreateLexer must return a pointer to the ILexer5 instance, GetLexerFactory must return a function that returns a pointer to the ILexer5 instance when called.
My conclusion: if Npp supports CreateLexer, there is no need for GetLexerFactory.
I hope I have not overlooked anything here.

Regarding GetLibraryPropertyNames and SetLibraryProperty, I don't really see the need, but who knows, maybe others see it differently.

GetNameSpace, on the other hand, is interesting if someone wants to replace an existing, builtin, lexer with another one.
For example, if I get the idea that my Python lexer would be much better than the one that Lexilla provides, Npp and I would have to use GetNameSpace so that it doesn't collide with the builtin lexer.

pryrt:
I updated the ILexer5 function list based on the comments from @Ekopalypse , and added descriptions of the functions that weren't also described for ILexer4.  (I think I got the prototype syntax similar to what had been used for ILexer4; feel free to edit and commit if I got it wrong.)

I also added a note that if lexer plugin developers have found a good reason to use the other functions, they should put in a GitHub issue requestion support.

content/docs/plugins.md
```diff
@@ -85,8 +85,54 @@ for any technical questions/answers and the announcement your new plugin.
...
+#### Notepad++ v8.4 and newer
+
+Notepad++ has transitioned to the [ILexer5](https://www.scintilla.org/LexillaDoc.html) interface as of Notepad++ v8.4.
+A lexer plugin needs to define all methods of this interface to ensure a smooth interaction.  
+
+The lexer functions to be exported by the plugin are
+- GetLexerCount  
+- GetLexerName  
+- CreateLexer
+- LexerNameFromID
```
donho, line 98: `LexerNameFromID` is deprecated in Leixilla, and not used and not needed absolutely in Notepad++.
Only `GetLexerCount`, `GetLexerName` and `CreateLexer` are mandatory (so far) for implementing an external lexer plugin for Notepad++.

content/docs/plugins.md
```diff
@@ -85,8 +85,54 @@ for any technical questions/answers and the announcement your new plugin.
...
+
+The [`SCI_SETILEXER`](https://www.scintilla.org/ScintillaDoc.html#SCI_SETILEXER) message is used to set the current lexer to your defined ILexer5 object.
```
donho, line 100: Calling `SCI_SETILEXER` is the job of Notepad++ to load external lexer plugin, but it's not necessary to use `SCI_SETILEXER` in the external lexer plugin.
I suggest to remove this line to avoid the confusion.

content/docs/plugins.md
```diff
@@ -85,8 +85,54 @@ for any technical questions/answers and the announcement your new plugin.
...
+2. LexerNameFromID
+~~~
+(typedef const char *(EXT_LEXER_DECL *LexerNameFromID(int identifier);)
+Args: identifier
+Return: Name of the lexer
+Description:
+	An optional function that returns the name for a lexer identifier.
+	This is a temporary affordance to make it easier to convert applications to using Lexilla. 
+	Applications should move to using lexer names instead of IDs. This function is deprecated.
+~~~
```
donho, line 125: To be removed.

content/docs/plugins.md
```diff
@@ -85,8 +85,54 @@ for any technical questions/answers and the announcement your new plugin.
...
+- GetLexerCount  
+- GetLexerName  
+- CreateLexer
+- LexerNameFromID
+
+The [`SCI_SETILEXER`](https://www.scintilla.org/ScintillaDoc.html#SCI_SETILEXER) message is used to set the current lexer to your defined ILexer5 object.
+
+The [Notepad++ Community Forum](https://community.notepad-plus-plus.org/) has a discussion on how to "[Make External Lexer Plugin work with v8.4](https://community.notepad-plus-plus.org/topic/22866/make-external-lexer-plugin-work-with-v8-4)".
+
+If you are starting with an ILexer4 lexer defined as below, to convert to ILexer5, it should work to simply add a `CreateLexer` function that returns the pointer to the ILexer5 interface.  In the future, Notepad++ may add support for the GetLibraryPropertyNames, SetLibraryProperty and GetNameSpace functions; if lexer-plugin developers find a good reason and want that support, they should [make an official feature request](https://github.com/notepad-plus-plus/notepad-plus-plus/issues).
```
donho, line 104: The external lexers done with an ILexer4 interface (with Scintilla 4.x headers), need to update their old Scintilla headers (Scintilla v4.x) to the new ones (Scintilla v5.x), and in additional add `CreateLexer`.
("PapyrusPlugin" has had upgrade to Scintilla5 before I adding `CreateLexer` in the plugin to make it work - I might be not clear on my commit comment on this part).

I suggest reword the above sentence to the following one:
*If you are starting with an ILexer4 lexer defined as below, to convert to ILexer5, you should update Scintilla headers to Scintilla5, and add a `CreateLexer` export function that returns the pointer to the ILexer5 interface.  In the future...*

donho (changes requested):
Please consider the suggested change.

pryrt:
updated per requested changes

donho (approved):

donho:
Thank you @pryrt  for the update!

PR #385, opened 2022-08-06
Views: add notes on view-rotation order 

pryrt:
(per [npp#8058](https://github.com/notepad-plus-plus/notepad-plus-plus…/issues/8058))

@alankilborn, does this look good?


alankilborn:
Nice addition.

PR #391, opened 2022-08-16
clarify Load Session for multi- vs single-instance

pryrt:


pryrt:
@alankilborn, does this make sense?

alankilborn:
> does this make sense?

Yes, good clarification.

PR #394, opened 2022-08-25
Clarify Shortcuts Clearing and Alt+X

### LINKED ISSUE #392: Expand upon how to remove a shortcut
Suggest changing existing wording:

`Use the Clear button to remove the existing shortcut for a given entry.`

to:

```
For non Scintilla commands, use the Clear button to remove an existing shortcut for the selected entry.  Alternatively, click the Modify button in the Shortcut mapper main window and then select None from the dropdown in the Shortcut window that appears, and choose OK.  

For a Scintilla commands shortcut the procedure to remove shortcut(s) from a function is a bit different.  Here the Clear key in the Shortcut mapper main window is constantly disabled and a current assignment must always be removed via a method started by pressing the Modify button.  Because Scintilla can have multiple keycombos for each command, all but one must be removed (using the Remove button), then the dropdown control set to None, and finally click Apply.
```

### LINKED ISSUE #393: Alt+x reference needs version qualifier
Text in existing manual:

`(Because the X on the right of the menu bar, which closes the active tab, was created as a menu action on the main menu bar with the name “X”, it’s accelerator is therefore Alt+X.)`

Note needs to be added that this only works in 8.4.2 and earlier versions; it has been removed in 8.4.3 and later.

Maybe add a suggestion for 8.4.3 to remap the _Close_ command for users that miss this capability?  [ref](https://community.notepad-plus-plus.org/topic/23404/where-did-alt-x-go-in-v8-4-4)

pryrt:
close #392
close #393


alankilborn (approved):

pryrt:
@alankilborn, does this work for those two issues?

PR #395, opened 2022-08-25
clarify plugins installation

pryrt:
- Per [this forum discussion](https://community.notepad-plus-plus.org/topic/23406/xml-tools-not-installed/), need to clarify permissions required
- while here, saw that the manual installation only showed the 32bit installation directory for plugins, which could cause confusion for 64bit installations or portable editions

### DIFF
--- a/content/docs/plugins.md
+++ b/content/docs/plugins.md
@@ -25,14 +25,20 @@ install, then select Install.
 ### Install plugin manually
 
 If the plugin you want to install is not listed in the Plugins Admin, you may
-still install it manually.  The plugin (in the DLL form) should be placed in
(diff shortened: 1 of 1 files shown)


alankilborn:
>Per [this forum discussion](https://community.notepad-plus-plus.org/topic/23406/xml-tools-not-installed/), need to clarify permissions required

That COULD be done, but maybe even better would be a Notepad++ code change alerting the user that the proper permissions do not allow the action?

pryrt:
That may be a good feature request for someone to submit.  However, given that there is user confusion over the way it is now, I figured I would document what _is_, rather than waiting for what _might be_.

PR #397, opened 2022-08-27
cursor -> caret

pryrt:
Change nearly all instances of "cursor" to "caret", except where it is actually talking about the mouse cursor (when hovering over text or GUI location, or when right-clicking).  Any place that "cursor" remains, made it clear it was talking about the mouse cursor.  

(... and kept **`NPPM_GETFILENAMEATCURSOR`**, as that is the name of the message.)


pryrt:
@alankilborn,

Did I change any that shouldn't have been changed?  Or should I have changed others.

alankilborn (approved):
I did my own independent search in the source code, and then compared hits from that with changes here.  I didn't see anything missing.

PR #400, opened 2022-09-07
Describe the config-file locations as per-user and machine-wide

### LINKED ISSUE #399: config file locations: machine vs per-user explanation
Alan suggested something akin to https://community.notepad-plus-plus.org/post/79576 be added to the User Manual.

pryrt:
Better explain the difference between the two, and when it will fall back to machine-wide closes #399

### DIFF
--- a/content/docs/config-files.md
+++ b/content/docs/config-files.md
@@ -8,30 +8,29 @@ weight: 115
 
 Notepad++ offers a comprehensive user interface to review or change most of its settings. However, there are some special cases where it is worthwhile to edit the configuration files directly, including:
 
-* Customizing the Context Menu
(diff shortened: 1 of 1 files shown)


pryrt:
@alankilborn, does this look like what you were hoping for?

alankilborn:
>does this look like what you were hoping for?

Looks good; I learned some stuff.
It was a bit hard to look thru the diff with all the whitespace changes.

alankilborn:
I found the setting for ignoring whitespace; then the change of the bulleting symbol stood out as changes.  At least that wasn't SO burdensome to see.

pryrt:
Sorry, I shouldn't've included the bullet and whitespace changes.  That was a bad decision on my part.

PR #427, opened 2022-11-07
change prolog encoding to UTF-8

### LINKED ISSUE #426: Auto-completion example encoding
Change to UTF-8

https://community.notepad-plus-plus.org/post/81194

pryrt:
closes #426


pryrt:
Experiments show that Unicode characters (either UTF-8 encoded or numeric entities) don't render properly.

PR #441, opened 2022-11-22
Update the Tab-to-Space note

### LINKED ISSUE #438: Clarify "tab to space" operation results
Can the manual be updated to clarify behavior of the _Edit_ > _Blank Operations_ > _TAB to Space_ command, to avoid the confusion that the petitioner of https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12528 had about its operation?

pryrt:
… to be more clear that "equivalent number of spaces" takes into account tab stops.

@alan-kilborn, please confirm this closes #438


alankilborn:
Is it worth adding?:

If you want a specific number of space characters to replace each tab character, you should use a replacement operation in Regular Expression search mode to replace `\t` with `\x20{`n`}`, where "n" is the number of spaces.

alankilborn (approved):

PR #483, opened 2023-02-20
clarify Preferences > Languages > Tab settings

pryrt:


pryrt:
@alankilborn, 

Does this phrasing work for you?

alankilborn (commented):
"If the text is replacing Tabs with spaces..." seems awkward.

alankilborn:
>"If the text is replacing Tabs with spaces..." seems awkward.

Suggested change:

If **_Notepad++_** is replacing Tabs with spaces, pressing the Tab key will insert enough literal spaces so that the next character will start at the beginning of the next tab stop.

pryrt:
Also, I will wait to see what happens with https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13186, in case that changes the wording needed here. ;-)

pryrt:
Since the "Replace by space" wording change wasn't included in the "adjacent" no-more-popup PR/commit, it may be a while before the wording gets changed (if ever). I don't want it waiting here indefinitely, so I'll do a separate fix to the manual if that changes in the future.

PR #485, opened 2023-02-22
update description of "volatile" search

pryrt:
The official spec for volatile vs normal select-and-find-next is enumerated here:

https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13145#issuecomment-1440292963

### DIFF
--- a/content/docs/searching.md
+++ b/content/docs/searching.md
@@ -306,27 +306,27 @@ The following commands, available through the Search menu or keyboard shortcuts,
     - The Search Results Window is created in response to any of the dialog-based **Find All** commands.
     - If such a search has been run in this instance of Notepad++, but the Search Results Window it is not visible, you can use  **Search > Search Results Window** to make it visible; if it is already visible, that command will switch the input focus between the Search Results Window and the current document.
 * **Select and Find Next / Select and Find Previous** Attempt to find the word the caret is in, or the selected text, down or up. The searched word or selection is remembered in the find history, and the search can be repeated with **Find Next / Find Previous**.
-    - The sequence of events is described below:
-        - all mentions of the _Find_ window in this sequence are still true even if the _Find_ window is not currently visible
-        - copies the selected text to _Find what_ box of _Find_ window, then uses it
+    - The specific search behavior:
+        - copies the selected text to _Find what_ box of _Find_ window for future use, then uses that same string for this specific search
             - if there is no selection and the caret is just at a single character position, it uses the word that the caret position is a part of (if any)
             - if the caret is in whitespace with no selection active, it will not perform a search, and the caret will remain where it was
-        - uses _Match case_ setting from _Find_ window
-        - uses _Match whole word only_ setting from _Find_ window
-        - uses _Wrap around_ setting from _Find_ window
-        - uses _Search mode_ = _Normal_
-        - selects and moves the caret to the next or previous match, using those search settings
+        - uses the following set of options:
+            - uses _Match case_ setting from _Find_ window
+            - uses _Match whole word only_ setting from _Find_ window
+            - uses _Wrap around_ setting from _Find_ window
+            - uses _Search mode_ = _Normal_ (regardless of its current setting in the _Find_ window)
+        - (all mentions of the _Find_ window in this search description are still true even if the _Find_ window is not currently visible)
 * **Find (volatile) Next / Find (volatile) Previous** Attempt to find the word the caret is in, or the selected text, down or up.
     - The searched word or selection is not remembered in the find history, and the search will not be repeatable with **Find Next / Find Previous**. That's why it's called volatile.  However, because it will have moved the caret and selection to the next match, repeated **Find (volatile) Next / Find (volatile) Previous** _will_ work as expected.
-    - The sequence of events is described below:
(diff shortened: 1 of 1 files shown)


pryrt:
@alankilborn, I slightly expanded on your text, to clarify regarding selection and caret-in-word and caret-in-whitespace.  Is it still an accurate description?

content/docs/searching.md
```diff
@@ -301,14 +301,32 @@ To some users these older results accumulating are an annoyance -- their data ma
...
+        - uses _Wrap around_ setting from _Find_ window
+        - uses _Search mode_ = _Normal_
+        - selects and moves the caret to the next or previous match, using those search settings
+* **Find (volatile) Next / Find (volatile) Previous** Attempt to find the word the caret is in, or the selected text, down or up.
+    - The searched word or selection is not remembered in the find history, and the search will not be repeatable with **Find Next / Find Previous**. That's why it's called volatile.  However, because it will have moved the caret and selection to the next match, repeated **Find (volatile) Next / Find (volatile) Previous** _will_ work as expected.
+    - The sequence of events is described below:
+        - uses the selected text as the search text
+            - if there is no selection and the caret is just at a single character position, it uses the word that the caret position is a part of (if any)
+            - if the caret is in whitespace with no selection active, it will not perform a search, and the caret will remain where it was
+        - uses _Match case_ = false (as this is most "flexible", i.e., "less strict")
```
alankilborn, line 325: Suggestion:
```
- uses the least-strict set of options, providing the most flexibility in a volatile search:
    - considers Match case to be unchecked
    - considers Match whole word only to be unchecked
    - considers Wrap around to be checked
    - considers Search mode to be Normal

File: content/docs/searching.md [hunk 2]
```diff
@@ -301,14 +301,32 @@ To some users these older results accumulating are an annoyance -- their data ma
...
-All dialog-free search actions respect the current search options set in the Find dialog like Match case or Wrap around.
+    - The sequence of events is described below:
+        - all mentions of the _Find_ window in this sequence are still true even if the _Find_ window is not currently visible
+        - copies the selected text to _Find what_ box of _Find_ window, then uses it
+            - if there is no selection and the caret is just at a single character position, it uses the word that the caret position is a part of (if any)
+            - if the caret is in whitespace with no selection active, it will not perform a search, and the caret will remain where it was
+        - uses _Match case_ setting from _Find_ window
+        - uses _Match whole word only_ setting from _Find_ window
+        - uses _Wrap around_ setting from _Find_ window
+        - uses _Search mode_ = _Normal_
```
alankilborn, line 317: Suggestion:  `-considers Search mode to be Normal` (regardless of its current setting in the Find window)

content/docs/searching.md
pryrt: done; I also balanced the description of non-volatile so that the options are an extra level of indentation, like here.

content/docs/searching.md
pryrt: done

alankilborn (approved):

NeilS-Wpg:
Some things to consider regarding the Searching section of the manual (current online version) that touch on the recent discussion:

Under "Dialog-free search/mark actions" / "Searching" there are four bullets. One of these is not at all like the other three. Bullet 2 refers to navigation within "Search results" and this is already well explained in the section dedicated to that topic a few screens up.

Also, the order of the other three bullets ought to go from most commonly used, to least commonly used. The volatile versions are the most sophisticated and least likely to be commonly used, so the order of the bullets could be revised to reflect this. Note that the expectation of what's most and least common is reflected in the number of the modifier keys used for the default (F3-centric) shortcuts: none, one, and two; I find this appropriate.

I noticed that in this same section, the text does not "hold the reader's hand" in that it does not include the default shortcuts. However, a few screens down in [Comparison between “Select and Find Next” and “Find Next (Volatile)”], the default shortcuts are included. It's helpful especially to new readers to see the shortcuts right where the action is described for the first time.

pryrt:
> Under "Dialog-free search/mark actions" / "Searching" there are four bullets. One of these is not at all like the other three. Bullet 2 refers to navigation within "Search results" and this is already well explained in the section dedicated to that topic a few screens up.

Good point.  That bullet doesn't need to be there at all.

> Also, the order of the other three bullets ought to go from most commonly used, to least commonly used. 

The new order, as per the merge of this PR, is "Next/Previous", "Select and Find Next/Previous", then "Find (volatile) Next/Previous" -- which matches the order in the menu system: when describing menu choices, I usually try to keep the order to match the sequence in the menu; but in this case, it happens to match your assumption that "volatile" is "least used".

> I noticed that in this same section, the text does not "hold the reader's hand" in that it does not include the default shortcuts. However, a few screens down in [Comparison between “Select and Find Next” and “Find Next (Volatile)”], the default shortcuts are included. It's helpful especially to new readers to see the shortcuts right where the action is described for the first time.

I hadn't remembered the "comparison between" was so far away from the actual description.  Now that the original description has more detail, it can probably be merged up into the main discussion of those features (including the keystrokes). Thanks for pointing that out.

I'll put it in my queue to remove the fourth bullet and to move/merge the "comparison between"

PR #486, opened 2023-02-23
update "comparison" section

pryrt:
- move it closer to the description of the non-volatile and volatile searches
- move the keystroke to the main description
- rework the comparison to make it more clear

per https://github.com/notepad-plus-plus/npp-usermanual/pull/485#issuecomment-1442011132

### DIFF
--- a/content/docs/searching.md
+++ b/content/docs/searching.md
@@ -302,7 +302,7 @@ The following commands, available through the Search menu or keyboard shortcuts,
 
 * **Find Next (`F3`)/ Find Previous (`Shift+F3`)**: Repeat searching the current search target, either down or up.
     - The "current search target" is whatever _Find what_ string was most-recently active from either the Find/Replace dialog or from the **Select and Find Next / Select and Find Previous**
-* **Select and Find Next (`Ctrl+F3`) / Select and Find Previous (`Ctrl+Shift+F3`)**: Search for the word the caret is in, or the selected text, down or up. The searched word or selection is remembered in the find history, and the search can be repeated with **Find Next / Find Previous**.
+* **Select and Find Next (`Ctrl+F3`) / Select and Find Previous (`Ctrl+Shift+F3`)**: Search for the word the caret is in, or the selected text, down or up. The searched word or selection is stored in the find history, and the search can be repeated with **Find Next / Find Previous**.
     - The specific search behavior:
         - copies the selected text to _Find what_ box of _Find_ window for future use, then uses that same string for this specific search
             - if there is no selection and the caret is just at a single character position, it uses the word that the caret position is a part of (if any)
@@ -314,7 +314,7 @@ The following commands, available through the Search menu or keyboard shortcuts,
             - uses _Search mode_ = _Normal_ (regardless of its current setting in the _Find_ window)
         - (all mentions of the _Find_ window in this search description are still true even if the _Find_ window is not currently visible)
 * **Find (Volatile) Next (`Ctrl+Alt+F3`) / Find (Volatile) Previous (`Ctrl+Alt+Shift+F3`)**: Search for the word the caret is in, or the selected text, down or up.
-    - The searched word or selection is not remembered in the find history, and the search will not be repeatable with **Find Next / Find Previous**. That's why it's called volatile.  However, because it will have moved the caret and selection to the next match, repeated **Find (volatile) Next / Find (volatile) Previous** _will_ work as expected.
+    - The searched word or selection is not stored in the find history, and the search will not be repeatable with **Find Next / Find Previous**. ("Volatile" here means "not stored".)  However, because it will have moved the caret and selection to the next match, repeated **Find (Volatile) Next / Find (Volatile) Previous** _will_ work as expected.
     - The specific search behavior:
         - uses the selected text as the search text, but does **not** overwrite the existing _Find what_ value in the _Find_ dialog
             - if there is no selection and the caret is just at a single character position, it uses the word that the caret position is a part of (if any)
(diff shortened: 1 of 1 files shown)


pryrt:
@NeilS-Wpg, does this change cover what you were asking for?

content/docs/searching.md
```diff
@@ -316,7 +313,7 @@ The following commands, available through the Search menu or keyboard shortcuts,
             - uses _Wrap around_ setting from _Find_ window
             - uses _Search mode_ = _Normal_ (regardless of its current setting in the _Find_ window)
```
NeilS-Wpg, line 314: Unless there was a recent change that I'm not aware of, this is incorrect. It's not the case in 8.1.9, 8.4.5 & 8.4.6. In fact (amusingly enough) only in the last few days did I discover the convenience of using this action when tuning or testing a regex: a single keyboard action launches a regex find, and delightfully, focus stays with the file tab. This is so nice I was thinking it deserves explicit mention, but probably not in this section (which should remain regex-innocent), but in the regex section.

alankilborn: @NeilS-Wpg It WAS a recent change.  This was never meant to take anything but literal text from a selection to do the find.  See #3229

NeilS-Wpg: Well, that's funny-sad, because when I realized I could do it (as I described it above) I felt like a kid at the beach who found money in the sand, and this glorious feature which has been around for years has just days ago been withdrawn... cosmically perhaps at the same instant that I discovered it.

It's not that I have a great dislike of working with the Find dialog, it's that I hate having to place focus back on the file after a find operation brings something of interest in view. (I believe I have a solution for this but it's likely not fully baked.)

Anyway, thanks (but NO THANKS!) for the info.

NeilS-Wpg:
@pryrt,

> does this change cover what you were asking for?

Yes, I think you have all the important things covered.

I did make a few notes while I was perusing, mostly on niggly stylistic points, but I'll share them anyways.

btw, it appears what's being edited is a working rev that's different from the published manual, so I'm a bit confused about published-oldish vs. new**ish** vs. new**est** changes. (Also I don't have lots of experience with github.)

----
Usage of the word _remembered_ --- that word is mostly associated with an action done by a human, and when speaking of machinery, it's nicer to say _retained_ or _stored_. A similar argument goes for _forgotten_ vs _discarded_. (And yes, I _can_ imagine a magnificent robot looking me straight in the eye and saying, "So, you don't think I remember things, am I understanding that correctly?")

----
> "That's why it's called volatile."

Stronger would be something like:
> In this context volatile means "not retained".

----
For the section:
> Comparison between "Select and Find Next" and "Find (Volatile) Next"
> "Both commands Select and Find Next and Find (Volatile) Next do almost the same thing: ..."

I'm thinking tighter wording would be:

> "Select and Find Next" and "Find (Volatile) Next" try to discover what to search for, and, they execute a search operation, in exactly the same way. However, the first can alter what a subsequent "Find Next" action will do, while the other leaves it untouched. Also, the first can "load" search target text into the "Find what" field of the Find Dialog, while the other leaves that untouched as well.

----
> (Note: The Settings > Preferences > Searching options might affect whether you see the Select and Find Next text the next time you open the Find dialog; but that action is storing the text in the Find what field, regardless.)

(For some reason I only saw this text in the "rich diff" view mode. where it's green; it's not visible in "source diff" mode.)

I find it confusing but I offer no alternate text as I'm not sure what you wanted to convey; I'm just suggesting review.

----
Here are two more things I noticed, though they're in sections not pertinent to the present Issue & Pull Request, and may even be ancient.

I saw:
> Finding characters in a specific range
> While regular expressions provide for specifying a range...

The section probably shouldn't start with a reference to an advanced topic, and most certainly not without stylizing the name of the topic (quotes, italics, Proper Case, link..) because that name, "regular expressions", to the uninitiated, is just two common words "everyone" knows the meaning of, but when combined like that make the mind reel in frustration.

I'd start with something like:
> It is sometimes useful to search for characters by their underlying (or "raw") numeric values, and moreover, to be able to specify a range of those numeric values... 

and then end by saying:

> For other ways of searching for characters by numeric value, see the later sections "Extended Search Mode" and "Regular Expressions". The latter supports specification of complex search criteria, such as multiple ranges for a single search operation.

(You may have had a good reason to mention that with regex it's "sometimes awkward when the start or end character isn’t easily typed in" so I'll leave it to you to decide if/where such a fragment ought appear.)

Side comment: in the published manual page `npp-user-manual.org/docs/searching/`, there are 48 occurrences of the phrase _regular expression_. Of those, 22 occur before the actual regex section, and many are lower case and not stylized. I'd urge you to pick a style and stick with it (for the early 22).

----
Finally, a super-mega-ultra triviality, and it's a complaint about _too much_ styling.

We see:
> Incremental search ...
> The ✗ allows you to close out...

The X in that text is **so** stylized that it causes confusion: "I see no Greek letters in the Incremental search pane! What am I missing?"

Far more clear would be: "The X button allows ..."

----
Assuming anyone's read this far, I hope it wasn't too annoying. If I had an inkling that I might end up writing this much stuff, I probably would've left things after the first sentence.

content/docs/searching.md
NeilS-Wpg: Maybe I'm not seeing the full picture here but I don't see much in the way of substantial discussion of the issue.

At https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3229 we have one user complaint in 2017, a year later another user posts in agreement, and a year after that, Scott joins in. And in saying:

> So for good or bad, it seems like these features (interactive Find, Select-and-Find, (keyboard)Find-next) might be inexorably tied together...

he essentially makes the point that "no matter what, you're gonna get burned", IOW, there will be times when these searches don't do what you were wanting.

The gold-plated full meal deal would be independent options pertaining to each subtype of search, and it's easy to imagine the np++ brain trust concluding it's not important enough to justify all the extra clutter.

So what prompted this very low intensity issue (with two "outside" participants in seven years, and being almost four years dormant) to get promoted to the point of being deemed worthy of action?

alankilborn: >...independent options pertaining to each subtype of search, and it's easy to imagine the np++ brain trust concluding it's not important enough to justify all the extra clutter.

Spot on!

>So what prompted this very low intensity issue ... to get promoted to the point of being deemed worthy of action?

This (very recent) user complaint: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13145

PR #502, opened 2023-03-22
add paragraph on \K-ReplaceAll

### LINKED ISSUE #499: \K requires REPLACE ALL
When reading [this post](https://community.notepad-plus-plus.org/post/85059), I was going to post a link to the line in the manual which informs users that if they use `\K` to reset the match, they need to use **Replace All**, not individual **Replace** commands.  Much to my chagrin, I couldn't find that admonition!  This must be rectified

pryrt:
closes #499


pryrt:
Not gonna hold my breath on that one. 😉

PR #507, opened 2023-04-24
Update Panel Description

### LINKED ISSUE #506: Possible additional info about "docked panels"
The information below might be good to revise/supplement the general info at the top of this section:  https://npp-user-manual.org/docs/views/#panels

There could be additional finer points to add...perhaps you can think of them?

Use or discard, at your option. :-)

----------------------------------------------------------------------------------

There are a variety of builtin "panels" (sub-windows) that can be used in Notepad++. These panels can be "docked" (so they appear as part of the main Notepad++ application window) on either the right, left, top, or bottom edges; or they can be made to "float", so they appear as a separate independent window. Docked panels can be moved easily to another edge, or to floating, by clicking on the panel's title and dragging it to the new location. Another way to make a docked panel float is to double-click its title bar; when made floating in this manner it may be returned to its previously docked location by double-clicking its title bar.

A panel docked to an edge of the Notepad++ main window may contain multiple functionalities. When it does, "tabs" identifying functionality will appear at the bottom of the docked panel window (contrast this with a panel with a single functionality -- the functionality is identified only in the title bar).

When dragging toward an edge of the Notepad++ main window that does not already contain a docked panel, when the mouse cursor gets close enough to the edge, the drag rectangle will "snap" to the edge, indicating that if the mouse is released, whatever is being dragged will "dock" in the location shown. However, when doing this with a floating panel, the drag rectangle will not "snap" if the given edge already contains a docked panel. To get it go join an existing docked panel, you have to end the drag over the title bar (at the top) in the case of a single docked item, or end the drag over the title bar OR the group of tabs (at the bottom) in the case of a multiple tabbed group panel.

With multiple tabs in a docked panel:

* double-click an individual tab: breaks only that tab out into a floating window (double-click the title bar of the newly created floating window to "put it back")

* double-click titlebar of any tab: breaks all tabs of that panel out into a floating window together (double-click a single tab of a floating group will put the single tab back in the previously docked location)

* to change the order of the tabs: drag and drop a tab within the other tabs (the tab to be moved will appear to the left of the tab dropped upon); note that for this type of move there is NO visual feedback to the user on what is occurring.

pryrt:
- used @alankilborn's suggestions, though I reworked the order
- added section on Panels getting too small to be visible
- link to the forum conversation about floating panels being left on non-visible monitors and fixing the Panel size through config-file editing (I will update the link to a FAQ if I ever get around to creating that FAQ)

closes #506


content/docs/views.md
```diff
@@ -195,17 +195,17 @@ The **Function List**[↗](#remembered-settings "Remembered Setting") toggle wil
...
+When dragging toward an edge of the Notepad++ main window that does not already contain a docked Panel, when the mouse cursor gets close enough to the edge, the drag rectangle will "snap" to the edge, indicating that if the mouse is released, whatever is being dragged will "dock" in the location shown. However, when doing this with a floating panel, the drag rectangle will not "snap" if the given edge already contains a docked panel. To get it go join an existing docked panel, you have to end the drag over the title bar (at the top) in the case of a single docked item, or end the drag over the title bar OR the group of tabs (at the bottom) in the case of a multiple tabbed group panel.
```
alankilborn, line 202: "To get it **go** join" -> "To get it **to** join"

PR #512, opened 2023-05-04
Update messages and notifications to include IDs

pryrt:
- add message ID to all NPPM messages
    - fix NPPM_DOCLISTDISABLEEXTCOLUMN name (should have been updated from NPPM_DOCLISTDISABLECOLUMN in v8.1.5)
    - add NPPM_DOCLISTDISABLEPATHCOLUMN (should have been added in v8.1.5)
- add notification ID to all NPPN notifications
    - fix spelling in NPPN descriptions
- updated the Message Key and Notification Key (the "key" of how each message or notification description is formatted)

Derived from @alankilborn's table in [this Community post](https://community.notepad-plus-plus.org/post/85989)


pryrt:
@alankilborn, is this a reasonable implementation of your chat-message suggestion?  Let me know if you have improvements.

alankilborn:
Seems reasonable to me.
I liked the "independent table" approach because I could copy that and turn it into code very quickly, but since I've done that myself I have it and no longer need it from a separate source.

PR #526, opened 2023-07-26
Update Notepad replacement instructions for Windows 11

whut:
Uninstalling UWP Notepad version does not make the replacement to be
respected. AFAIK this is because there is `UseFilter` set, with which
only the exact full paths mentioned in registry subkeys are used for
replacement. So either UseFilter must be set to 0, or (IMO better) just
follow the approach taken by UWP Notepad and add Debugger to each
subkey.

Still then UWP Notepad would be run from Start Menu and when
double-clicking on text file. As UWP apps are activated in special way,
not just by running a binary, this registry trick does not work for
them. So to fix ability to run from Start Menu UWP Notepad must just be
uninstalled and then sadly you have no Notepad shortcut in Start Menu.
After uninstall you will observe that there is no application associated
with text files, not even built-in Notepad. This is because there is
another registry setting that disables using built-in Notepad to open
files, so it also must be removed.

### DIFF
--- a/content/docs/other-resources.md
+++ b/content/docs/other-resources.md
@@ -3,8 +3,8 @@ title: Other Resources
 weight: 160
 ---
 
-
 ## Notepad Replacement
+
 Notepad is a default text editor shipped with Windows. You may want to use Notepad++ instead of Notepad. However, there's no obvious way to do it.
 From the version 7.5.9 onward, you can run the following command to make Notepad++ replace Notepad (run in `cmd.exe` with Administrator privileges)[†](#registry-edit-warning):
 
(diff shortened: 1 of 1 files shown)


content/docs/other-resources.md
```diff
@@ -3,30 +3,34 @@ title: Other Resources
 weight: 160
 ---
 
-
 ## Notepad Replacement
+
 Notepad is a default text editor shipped with Windows. You may want to use Notepad++ instead of Notepad. However, there's no obvious way to do it.
-From the version 7.5.9 onward, you can run the following command to make Notepad++ replace Notepad (run in `cmd.exe` with Administrator privileges)[†](#registry-edit-warning):
+From the version 7.5.9 onward, you can run the following command to make Notepad++ replace Notepad (run in `cmd.exe` with Administrator privileges[†](#registry-edit-warning)):
```
pryrt, line 9: The footnote applies to the whole sentence, not just the parenthetical, which is why it was outside the parentheses.  Do you have a reason for changing this?

content/docs/other-resources.md
```diff
@@ -3,30 +3,34 @@ title: Other Resources
...
 
-To remove UWP version of Notepad, you have to launch powershell with admin privileges, then run the following command:
-```powershell
-Get-AppxPackage *Microsoft.WindowsNotepad* | Remove-AppxPackage
+```batch
+reg delete "HKLM\Software\Microsoft\Windows NT\CurrentVersion\Image File Execution Options\notepad.exe" /v "Debugger" /f
 ```
 
-For restoring Notepad, launch "Microsoft Store" via Start menu of Windows, search "Notepad", select "Windows Notepad", then click on "Install" button.
+Additionally on Windows 11 reinstall Notepad (launch "Microsoft Store" via Start menu of Windows, search "Notepad", select "Windows Notepad", then click on "Install" button) and run:
```
pryrt, line 29: Please rephrase.  I suggest something like:

> On Windows 11, you must additionally reinstall Notepad (launch "Microsoft Store" via Start menu of Windows, search "Notepad", select "Windows Notepad", then click on "Install" button) and then run:

pryrt (changes requested):

content/docs/other-resources.md
whut: I though it is more connected with what is in parenthesis, but you are right, it ifor the whole sentence, reverted.

content/docs/other-resources.md
whut: Thanks for styling comments, appreciate that!

In the end I changed the instructions again, as I finally understood what exactly is done there in registry for UWP Notepad: UWP Notepad use `UseFilter` in Image File Execution Options, which allows you to match executables using full paths (in conjunction with `FilterFullPath` in subkey), which avoids the (low) risk of replacing some unrelated binary with notepad.exe as name

whut:
I had to update instructions again, as I noticed that UWP Notepad uninstall is still necessary as without it it will be used when double-clicking text files or from Start Menu. And to be able to use legacy Notepad to open text files (now redirected to Notepad++) yet another registry setting must be changed. Don't hesitate if there are any styling improvements.

pryrt:
I made a slight tweak: I thought it would be easier to do it myself in the GitHub web interface than ask you to do it.  However, it apparently needs a "force push" and I cannot do that from the web.  Can you do a force-push of the commit I made in your branch?  Or am I going to have to take it to a command line client and figure it out myself?

pryrt (approved):
it's ready

PR #533, opened 2023-07-28
Getting Started: Installer-vs-Portable

### LINKED ISSUE #532: GettingStarted: Clarify differences between installation methods
per  @victorel-petrovich and @mkupper in [Community Forum](https://community.notepad-plus-plus.org/post/88181)

pryrt:
Clarifications made in this update:
- installer usually needs admin privileges
- installer can install to other directories than just "Program Files"
- installer can choose to not use %APPDATA%
- installer includes Shell Extension
- portable uses local configuration
- portable does not set context-menu action
- differentiate mini/minimalist compared to portable
- add section on "Program Files" restrictions

closes #532

### DIFF
--- a/content/docs/getting-started.md
+++ b/content/docs/getting-started.md
@@ -5,24 +5,48 @@ weight: 1
 ---
 
 ## What is Notepad++
+
 Notepad++ is a text editor and source code editor for use under Microsoft Windows. It supports around 80 programming languages with syntax highlighting and code folding. It allows working with multiple open files in a single window, thanks to its tabbed editing interface.
 Notepad++ is available under [GPL](http://www.gnu.org/licenses/gpl-3.0.html) and distributed as [free software](https://www.fsf.org/).
 
 You may visit the Notepad++ website at https://notepad-plus-plus.org/
 
 ## Download Notepad++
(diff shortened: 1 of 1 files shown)


victorel-petrovich:
Good, now I won't need to keep my own little .txt file detailing all that :)

I just read your changes and here are a few remarks:

- at "## Install Notepad++ using the installer", will need admin rights not only at program installation, but also later when installing plugins (in my limited experience, at least in case of NppExec and PythonScript ). If choose to install in a directory other than the standard "Program Files", then it will still require admin rights (for example, I tried creating a folder "MyProgrFiles" in my local User directory, and that was the case). 
- at both Installer and Portable subsections: that right-click menu entry (gotten with shell extension or .reg method), they call it now "Edit with Notepad++" 
- at mentioning the differences for the mini-portable, what it doesn't include: I'd mention no plugins / no plugin-capability at all ; new users might not know what "Plugins Admin" means.

pryrt:
> * no plugin-capability at all

Actually, if you manually install a plugin, the minimalist edition will use the plugin just fine.  It just doesn't come with any of the default plugins

PR #535, opened 2023-07-29
Getting Started: Further updates/clarifications

pryrt:
- Remove phrase about plugins requiring elevated permissions; that's already documented in plugins.md#permissions
- Add a phrase that the default config-file location is AppData before describing the "Don't use %APPDATA%" option
- Include a note that "Edit with Notepad++" may be phrased differently depending on version and localization at the first mention of "Edit with Notepad++".  The reader is assumed to be able to understand that the same note applies to the other mentions of "Edit with Notepad++"
- Move the paragraph on installer options to the _end_ of the installer section
- Add a sentence that Notepad++ can be uninstalled using standard OS uninstall procedures

Portable section:
- Give an introductory paragraph that explains that everything goes in the same folder, with no associations or context-menu entries
- Make two simple, one-to-two sentence paragraphs to describe the the contents of the the normal portable vs mini-portable
- Add a couple sentences on "uninstalling" the portable

Restrictions:
- rework the paragraph on doLocalConf not working in "Program Files" now that we know it was intentional.

### DIFF
--- a/content/docs/getting-started.md
+++ b/content/docs/getting-started.md
@@ -23,13 +23,15 @@ Determine whether your system requires the 64-bit or ARM 64-bit or 32-bit build 
 1. Download the installer
 2. Run the executable binary and follow the installation flow
 
+The installer will likely require Administrative privileges in order to install Notepad++.  If you do not have Administrative privileges, you either need to find someone who does, or tell the installer to use a location where you _do_ have write permission (though that may still ask for Administrator privileges), or run a portable edition from a directory where you have write permission.
+
+By default, the installer will use `%AppData%\Notepad++` to [store your settings](../config-files/#configuration-files-location); but when you use the installer to install to a directory other than "Program Files", you can choose the installation option "Don't use %APPDATA%" to instead use a "[local configuration](../config-files/#configuration-files-location)".
+
+The installer should also install the [Shell Extension](../shell-extension/) to add **Edit with Notepad++** to the Windows Right-Click context menu.  (Your specific version or translation may use a different phrasing for that entry.)
+
 If you are doing managed installation or otherwise want to control the installer from the command line, the installer has a few [command line options](../command-prompt/#installer-options).
 
-The installer will likely require Administrative privileges in order to install Notepad++ in Windows's standard "Program Files" location, and to install/update plugins when installed in that location.  If you do not have Administrative privileges, you either need to find someone who does, or tell the installer to use a location where you _do_ have write permission (though that may still ask for Administrator privileges), or run a portable edition from a directory where you have write permission.
-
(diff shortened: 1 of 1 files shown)


victorel-petrovich:
IMHO, you may want to wait a bit before merging or closing, if you want people to give feedback. 

> Remove phrase about plugins requiring elevated permissions; that's already documented in plugins.md#permissions 

:/ , but we already discussed this perspective.

victorel-petrovich:
Needing or not admin rights is an important difference between Installer and Portable, and this concerns not only installation but further usage (thus plugins).

pryrt:
> you may want to wait a bit before merging or closing

Sorry, I had read this statement in the issue:
> Well, someone's editings will never be perfect by me or any other particular editor, but I guess that's how open collaboration works.

That led me to believe you had resigned yourself to our difference of opinion on the remaining things you disagreed with.

What if I change the first line of the "Program Files Restrictions":
> Windows is particular about what goes into "Program Files", and restricts permissions there, usually requiring Administrator Privilege to install or edit there -- this includes installing or removing plugins.

I am willing to put it there, but not in the Installer section, because the admin requirement for plugins is a result of the location and the OS rules, _not_ of Notepad++ itself or whether that directory populated with installer or zipfile.

victorel-petrovich:
> That led me to believe you had resigned yourself to our difference of opinion on the remaining things you disagreed with. 

That was true for the things we discussed on that issue's page. But about "Remove phrase about plugins requiring elevated permissions; that's already documented in plugins.md#permissions ", I only found out later, on this page of commit. 

Including it in "Program Files Restrictions", will be okay / better than nothing. 
However:

> the admin requirement for plugins is a result of the location and the OS rules, not of Notepad++ itself or whether that directory populated with installer or zipfile. 

That contradicts my experience: I tried installing in C:\Users\Vic\MyOwnProgrFiles , and it still asked admin pass later when installing plugins (at least in case of NppExec that I tried then). 

So it appears to be about installer vs zip, not just location.

victorel-petrovich:
I tested now (installing in C:\Users\Vic\MyOwnProgrFiles), and I found that it actually depends on whether, during installation, I checked or not the box: "Don't use %AppData%" . 
If I don't check (let it use %AppData%), then it will ask for admin pass later at plugin install. That's what I must have done earlier.
If I check it, then it will not ask. 

Weird, isn't it?

So the answer would have to be more nuanced: If you use installer, and either choose to install in Program Files, or, during installation, leave unchecked  "Don't use %AppData%", then you'll also need admin right for later installing plugins (via Plugins Admin).

victorel-petrovich:
If you ask me, there are *other* things that you could leave out of that Installer section, that most users that care to use such an editor already find trivial, like:
```
1. Download the installer
2. Run the executable binary and follow the installation flow

[...]  you either need to find someone who does, [...]  // what to do if you don't have admin rights
```

pryrt:
I disagree with you, but one phrase was not something that's worth arguing over: hopefully #536 will be sufficient for you.

victorel-petrovich:
I agree, I should not have spent so much time typing all the above.

PR #545, opened 2023-08-15
v8.5.6: NPPM_ALLOCATEINDICATOR

### LINKED ISSUE #542: Add info about NPPM_ALLOCATEINDICATOR
With https://github.com/notepad-plus-plus/notepad-plus-plus/commit/de25873cb3352ee59d883e95e80c91806944e348 and presumably release of Notepad++ 8.5.6, it will be possible for Notepad++ to allocate indicators similar to how it does markers and command-ids.

This should be documented in the plugin command section of the user manual; suggested text is something like this:

--------------------------------

[2137] NPPM_ALLOCATEINDICATOR
Obtains a number of consecutive indicator IDs dynamically, with the guarantee of these IDs not clashing with Notepad++'s own internal indicator usage, nor usage by any other plugins that use this API to allocate their needed indicators.

Parameters:

wParam [in] - int numberOfIndicators
lParam [out] - int * firstIndicatorID
Return value - Returns 0 on failure, nonzero on success

If successful, (*firstIndicatorID)+0 thru (*firstIndicatorID)+numberOfIndicators-1 are the indicator numbers that have been set as reserved for use by the caller.

pryrt:
closes #542

### DIFF
--- a/content/docs/plugin-communication.md
+++ b/content/docs/plugin-communication.md
@@ -166,9 +166,8 @@ struct toolbarIconsWithDarkMode {
 **Return value**
 : Returns 0 on failure, nonzero on success
 
-**Other information**:
-
-If successful, `(*firstIndicatorID)+0` thru `(*firstIndicatorID)+numberOfIndicators-1` are the indicator numbers that have been set as reserved for use by the caller.
+**Other information**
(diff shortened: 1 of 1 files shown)


pryrt:
@alankilborn, does this look good to you?

alankilborn (approved):

PR #551, opened 2023-08-25
clarify Increase/Decrease Line Indent

pryrt:
per Alan's chat message


pryrt:
@alankilborn,

does this look sufficient to you?

PR #562, opened 2023-09-18
minor amendments 2

abcbuck:
updating searching.md up to '### Search results window' (lines 104-193)

changes:
- Better explain `Alt` usage
- `Code block format` `inputs`
- Improve wording
- Include 'Settings' in link to settings
- Remove a question directed to nobody in particular
- Add link to Project Panels in 'Find in Projects tab' section

### DIFF
--- a/content/docs/searching.md
+++ b/content/docs/searching.md
@@ -145,7 +145,7 @@ If users need to exclude folders with the same name (or names matching the speci
     * `!\bin*` will not search any files in the `bin` folder or `bin64` folder (or any other directory that matches `bin*`),
     * `!+\log*` will _recursively_ not search any files in folders that start with log (so directories like `.\log`, `.\logs`, `.\other\logfiles`, `.\many\layers\deep\log` will all be excluded from the search).
 
-    *Note*: Inclusion of subfolders of excluded folders is not allowed, and any such pattern will be ignored.
+    *Note*: There is no opposite ("inclusive") version of the folder-exclusion syntax.  This means that other than checking **☐ In all sub-folders**, you cannot include specific folders in the search.  In particular, you cannot include specific subfolders of excluded folders (`!\skip \skip\exceptThis` will _not_ work to skip the `\skip` subfolder but to still search the `\skip\exceptThis` subfolder).
 
 * As of Notepad++ v8.2, if you hover your mouse cursor over the **Filters** label, a helpful popup will show example syntax to you.
 * Please also note that the PathMatchSpec() Windows API is being used for the **Filters**, as its behavior departs from cmd.exe wildcard parsing sometimes.


content/docs/searching.md
```diff
@@ -101,29 +101,30 @@ The various action buttons available include:
...
+The above actions may be initiated via mouse by pressing the appropriate button, or via special `Alt` key combinations.  Notepad++ underlines a single character in the text of *most* of the buttons.  Pressing `Alt` and one of the underlined characters will be the same as pressing the same button with the mouse, i.e., the chosen action will be initiated.  The `Alt` technique works for controls other than buttons as well, e.g., a checkbox control can be checked/unchecked via pressing its `Alt` key command.  In a similar vein, pressing `Alt` outside one of the Find dialog windows underlines items in the top main menu.
```
pryrt, line 106: Your "correction" isn't entirely accurate.  But neither is the original phraseology.

There are times when bringing up one of the search menus does **not** show any of the underlines on one of the buttons: when that happens, clicking `Alt` alone _will_ show the missing underlines.  But the missing-underline-condition doesn't happen every time, which is what the original phrasing implied.

A better phrasing would be:
> Notepad++ can underline a single character in the text of *most* of the buttons (oftentimes, it is automatic; however, if you ever cannot see the underlines, then just click the `Alt` key, and they will appear).

You also added:
> In a similar vein, pressing `Alt` outside one of the Find dialog windows underlines items in the top main menu.

That is _not_ the point of the original comment about press-and-release the `Alt` key, and does not belong in this page of the manual.  (And since it's common Windows behavior, doesn't really need to be in the user manual at all.)

content/docs/searching.md
```diff
@@ -101,29 +101,30 @@ The various action buttons available include:
...
+Notepad++ uses a flashing of the Find dialog window and the main Notepad++ window itself (when the Find dialog is not open) to indicate that search text has not been found (or possibly that a **Wrap around** in the search has occurred).  In general, if a  in no matches, and the Find dialog window is open, that window will flash briefly as a failure indication.  If the Find dialog window is NOT open, and a failed search is initiated (e.g. via **Find Next** on the **Search** menu), the main Notepad++ window will flash briefly, again, as an indicator of the lack of success.  With the Find dialog window closed, but with **Wrap around** previously activated, a search that causes a wrap at an end of the file to occur will also cause the Notepad++ main window to flash.  In addition, audible feedback will be provided if a **Find Next** or **Replace** action results in the **Find what** text not being encountered; the sound can be muted using the **Mute all sounds** option in [**Settings > Preferences > MISC.**](../preferences/#misc).
```
pryrt, line 125: > "if a  in no"

I'm not sure why you didn't like the original; "if a search results in no matches", which was a perfectly valid clause (search being used as a noun, and results being used as a verb), but if you don't like it and have a better phrasing, that's fine.  But "if a in no" ain't it. ;-)

content/docs/searching.md
```diff
@@ -139,30 +140,30 @@ The **Filters** list is a space-separated list of wildcard expressions that cmd.
...
-    * `!\tests` will not search any files in the `tests` folder
-    * `!\bin*` will not search any files in the `bin` folder or `bin64` folder (or any other directory that matches `bin*`)
-    * `!+\log*` will _recursively_ not search any files in folders that start with log (so directories like `.\log`, `.\logs`, `.\other\logfiles`, `.\many\layers\deep\log` will all be excluded from the search)
+If users need to exclude folders with the same name (or names matching the specific pattern) in all levels, the `+` should be put between `!` and `\` to exclude them recursively. For example:
+    * `!\tests` will not search any files in the `tests` folder,
+    * `!\bin*` will not search any files in the `bin` folder or `bin64` folder (or any other directory that matches `bin*`),
+    * `!+\log*` will _recursively_ not search any files in folders that start with log (so directories like `.\log`, `.\logs`, `.\other\logfiles`, `.\many\layers\deep\log` will all be excluded from the search).
 
-    *Note*: "inclusion of folder" is not allowed, and such pattern will be ignored
+    *Note*: Inclusion of subfolders of excluded folders is not allowed, and any such pattern will be ignored.
```
pryrt, line 148: No.  There is no way to use the filter to include subfolders.
The syntax `!\tests` is valid to exclude a subfolder, but `\tests` will _not_ cause a new folder to be included, which was the point of that statement.  By saying "of excluded folders", it implies that you are trying to undo a previous exclusion-filter.

content/docs/searching.md
```diff
@@ -139,30 +140,30 @@ The **Filters** list is a space-separated list of wildcard expressions that cmd.
...
 
-* As of Notepad++ v8.2, if you hover your mouse cursor over the **Filters:** label, a helpful popup will show example syntax for you
+* As of Notepad++ v8.2, if you hover your mouse cursor over the **Filters** label, a helpful popup will show example syntax to you.
 * Please also note that the PathMatchSpec() Windows API is being used for the **Filters**, as its behavior departs from cmd.exe wildcard parsing sometimes.
-* To find only files that have no extension, you cannot just say `*.`, because of quirks in the PathMatchSpec().  Instead, you can search for any file that _doesn't_ have at least one letter in the extension: `!*.?*`
+* To find only files that have no extension, you cannot just say `*.`, because this would still include a `.`.  Instead, you can search for any file that _doesn't_ have at least one letter in the extension: `!*.?*`.
```
pryrt, line 152: > this would still include a `.`.

That's not technically the reason.  Try this experiment at the command line: create a file named `a` with no extension, then do `dir *.` -- you will notice that it finds `a` even though it doesn't have the dot.  

Maybe a better phrasing would be:
> you cannot just say `*.` (despite that filter working in a Windows command prompt)

Or just not give a reason-clause, if you think that any attempted clarification will confuse readers more than it clarifies.

pryrt (changes requested):
See individual comments

content/docs/searching.md
abcbuck: I see. I wonder how at times it works and then it doesn't. It seems to work consistently on my end.
Copied your improvement. Deleted the last part which I thought was what was meant.

content/docs/searching.md
abcbuck: Funny. :)
I used the search to find some "search results" and must have activated the replacement/deletion feature by pressing `Enter`. I thought I had fixed it already but I must have pressed it twice.

content/docs/searching.md
abcbuck: Isn't this how you would use an inclusion statement, though? You start with everything from a given directory and exclude folders. The only folders you could include then are subfolders of excluded folders.

I was confused as I read this because this is about *filters* whose only point is to exclude things, so I thought inclusion would be about exactly that: partially reversing exclusions, or in other words: excluding everything but a part of a directory.
If, to go with your example, `!\tests` isn't part of the filter, `\tests` is never needed as it is included in the search like every other folder in the directory. However, `\tests\subfolder` would make some sense to reinclude an exiled part.

Inclusion of folders isn't allowed, correct, that's what the original line stated, and I got that. However, as folders by themselves are never in need of being included, I think it would make sense to write explicitly of subfolders.

content/docs/searching.md
abcbuck: This `.` goes to you.

content/docs/searching.md
pryrt: > this is about filters whose only point is to exclude things
 
I guess that's not quite the way I think of filters. Because the filter `*.cpp` is about what to include, rather than what to exclude -- though I guess the implication is "exclude everything, then just include `*.cpp`"

> The only folders you could include then are subfolders of excluded folders.

Unless you're non-recursive.  (Though again, I guess you'd might think of it as "non-recursive is essentially excluding all subdirectories")

Maybe a different wording:
```
*Note*: There is no opposite ("inclusive") version of the folder-exclusion syntax.  This means that if your search is not using `☐ In all sub-folders`, you cannot add a specific subfolder to include; it also means that if you exclude a subfolder, you cannot tell it to include a specific deeper-subfolder (so `!\skip \skip\exceptThis` will _not_ work to skip the `\skip` subfolder but to still search the `\skip\exceptThis` deeper-subfolder).
```

PR #563, opened 2023-09-18
minor amendments 3

abcbuck:
updating searching.md up to '## Dialog-free search/mark actions' (lines 194-299)

changes:
- `Code block format` `inputs`
- Correct search results example
- Add link to searching preferences
- Use present tense instead of future in writing
- Remove unnecessary CAPS highlighting
- Improve wording
- Delete redundant paragraphs
- Use 'fold' instead of 'collapse' as this is the current Notepad++ term
- Remove questions expressing uncertainty as to the intentions of the Notepad++ creators by an editor writing the manual

### DIFF
--- a/content/docs/searching.md
+++ b/content/docs/searching.md
@@ -192,7 +192,7 @@ Bookmarks, whether visible or not, have a menu that can control and navigate Boo
 
 ### Search results window
 
-After running one or more **Find All in ...** commands, a new **Search results** window appears, and within it is placed a **Search results** tab.  The **Search results** window may be opened and/or given input focus by using the menu command **Search > Search Results Window** or the F7 keyboard shortcut.  *Note:*  That menu command will seem to not do anything if there haven't been any **Find All in ...** commands run since Notepad++ was opened.
+After running one or more **Find All in ...** commands, a new **Search results** window appears, and within it is placed a **Search results** tab.  The **Search results** window may be opened and/or given input focus by using the menu command **Search > Search Results Window** or the `F7` keyboard shortcut.  *Note:*  That menu command will seem to not do anything if there haven't been any **Find All in ...** commands run since Notepad++ was opened.
 
 *Definition:*  **Find All in ...** commands include:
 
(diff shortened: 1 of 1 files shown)


content/docs/searching.md
```diff
@@ -207,98 +207,94 @@ The **Search results** window by default appears docked at the bottom of the Not
...
+There is a capability to copy the list of files containing hits from past searches (v8.0.0 and later).  The **Copy Pathname(s)** context menu command copies to the clipboard the full pathnames of all files appearing in **Search results**.
 
 ##### Other commands
 
 There are some commands that don't need a lot of explanation; these are:
 
-* **Collapse all**
-* **Uncollapse all**
+* **Fold all**
+* **Unfold all**
```
pryrt, line 256: Please rephrase
> * **Fold all** (previously called **Collapse all** before v8.4.6)
> * **Unfold all** (previously called **Collapse all** before v8.4.6)

When behavior has changed in "recent" versions, we need to describe both behaviors, because this manual is valid for all recent versions.  (Currently, "recent" means v8.0 or newer.)

content/docs/searching.md
```diff
@@ -207,98 +207,94 @@ The **Search results** window by default appears docked at the bottom of the Not
...
 
 ##### Searching in previously-found results (secondary searching)
 
-Perhaps you have done a search and your results are in a tab in the **Search results** window.  Now you'd like to conduct a search but with a scope of only the files that have previous matches.  Or maybe you want to look only in the *lines* matched by previous searches, not only the matched files, tightening the search criteria even more.  Can you do this sort of second-level searching with Notepad++?  Yes, by *RightClick*ing the **Search results** window client area and selecting **Find in these search results...**.
+Assume, you have done a search and your results are in a tab in the **Search results** window.  Now you'd like to conduct a search but with a scope of only the files that have previous matches.  Or maybe you want to look only in the *lines* matched by previous searches, not only the matched files, tightening the search criteria even more.  Can you do this sort of second-level searching with Notepad++?  Yes, by `RightClick`ing the **Search results** window client area and selecting **Find in these search results...**.
```
pryrt, line 273: > Assume, you have done

I think it would be better:
> Assume that you have done

pryrt (changes requested):
See comments

content/docs/searching.md
abcbuck: Noted.

content/docs/searching.md
abcbuck: Assimilated.

content/docs/searching.md
pryrt: > Oh, also, how about we delete the sentence above that? Say these items don't need a lot of explanation and then proceed to explain them seems contradictory...

Good idea.

PR #570, opened 2023-09-21
Amend session.md

abcbuck:
- Capitalize sentence beginnings
- Update menu item name to reflect current Notepad++ usage

I looked for changes in the patch notes since v8.5 regarding name changes of menu items but didn't find anything. Is there a quick way to check this other than calling each site with a new patch (like https://notepad-plus-plus.org/downloads/v8.5.7/)?


pryrt:
> Is there a quick way to check this other than calling each site with a new patch

https://github.com/notepad-plus-plus/notepad-plus-plus/wiki/Changes

PR #612, opened 2023-12-20
replicate Don's Multi-Edit tutorial to editing.md

pryrt:
<https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14509>


alankilborn:
Looks good.

PR #621, opened 2024-01-24
backward regex surprise example

pryrt:
don't make people search for "sometimes surprising results" -- extract the example from the original issue 3640


alankilborn:
Nicely done.

PR #626, opened 2024-02-02
Add Note on "Match whole word"

pryrt:
The implications of "match whole word" aren't always immediately obvious.  So make it clear that selecting multiple words, or mixed words and punctuation may not behave as expected when the "match whole word" option is active.

(Adding this Note was prompted by the experiments I ran when replying to [this forum post](https://community.notepad-plus-plus.org/topic/25444/all-occurrences-of-token-not-all-the-same-texts-are-marked-highlighted).)


pryrt:
@alankilborn, does my proposed Note make sense to you?  Do you think it needs more (or less) detail?  Other comments?

PR #629, opened 2024-02-04
Add Document Switcher docs to new "User Interface" page

### LINKED ISSUE #628: Improve Document Switcher documentation
The current extent of the Document Switcher documentation seems to be:

_☐ Enable: Hitting Ctrl+TAB will allow you to easily switch through all the open documents._

Some suggested changes:

"Hitting" sounds rather violent; perhaps it could be "Pressing"?

Also, it needs further discussion, on these points (ideas expressed are rather ad-hoc, and probably need the user manual moderator's spin on the wording):

* The idea behind usage is that one would hold the Ctrl key and then press+release the Tab key.  This gets the Doc Switcher on screen in the form of a yellow-backgrounded popup showing the current tab set for the view (or a subset if there are too many tabs).  Initially the name of the tab to the right of the currently-active tab will be bolded, with boldness indicating which tab will be made activate when all keys have been released.  With the Doc Switcher on screen, repeated press+release of the Tab key will move the bold highlight downward on screen to the next tab name in the list.  When the desired tab appears bolded, release all keys to make the bolded tab active in the editor.

* If the bottom-most entry is reached and another Tab press+release is executed, the topmost entry will be bolded, i.e., the highlight will wrap around to the first entry in the list (or, if the list was too long to entirely be displayed, the next entry will be made visible and will become bold.

* Adding Shift in such that both Ctrl and Shift are the held keys will cause the bold highlight to move to the previous list entry rather than the next.

* The Doc Switcher functionality can also be achieved using the mouse alone:  Right-click in the editing area for a tab and hold it, then begin scrolling the mouse wheel (in either direction).

pryrt:
closes #628 

Also moved the sections under Cheat Sheet into the new User Interface page instead (leaving a link from Other Resources > Cheat Sheet to the new page)


content/docs/user-interface.md
```diff
@@ -0,0 +1,121 @@
...
+- **Generate from Selection into Clipboard** - Will use the text that is currently selected, calculate the hash, and put the hash results in the Clipboard.
+
+## Document Switcher
+
+The Document Switcher feature (which can be enabled or disabled using the **[Settings > Preferences > MISC](../preferences/#misc) > Document Switcher** settings) can be used to quickly switch between all the open files in the active View.
+
+When you hold the <kbd>Ctrl</kbd> key and then press and release the <kbd>Tab</kbd> key, the Document Switcher appears as a popup with a yellow background, showing the current list of tabs for the active View (or a partial list, if there are too many tabs); the popup will remain until you release the <kbd>Ctrl</kbd> key.  One tab in the list will be selected (displayed in bold), indicating which tab will be made activate when all keys have been released.  With the Document Switcher on screen, repeated press and release of the <kbd>Tab</kbd> key will move the selection (bold highlight) downward on the screen, to the next tab name in the list.  When the desired tab is selected (bold), release all keys to activate that selected tab.
```
alankilborn, line 111: "showing the current list of tabs for the active View" <-- actually, it combines the files from both views; in MRU mode they are likely interleaved, in non-MRU mode the view1 tabs are at the top of the list, the view2 tabs at the bottom

content/docs/user-interface.md
```diff
@@ -0,0 +1,121 @@
...
+
+If you hold down <kbd>Shift</kbd> (while still holding the <kbd>Ctrl</kbd> key to keep you in Document Switcher mode), and then press and release the <kbd>Tab</kbd>, it will cause the selection (bold highlight) to move to the previous entry in the tab list, rather than the next.  (And in the long-list scenario, it would scroll to the next tab above the top in the list, making that visible and scrolling the bottom tab in the list out of the visible popup.)
+
+The Document Switcher functionality can also be achieved using just the mouse (if you have a scroll wheel): Right-click in the editing area for a tab and hold the right mouse button, then begin scrolling the mouse wheel (in either direction) to display the Document Switcher popup; further scrolling of the scroll wheel will change which tab is selected in the list.  Releasing the right mouse button will cause the currently-selected tab to be activated in the current View.
```
alankilborn, line 119: Additional mouse functionality:  while still holding down the right button, if you move the mouse and point to an entry somewhere else in the Doc Switcher list, and then left-click, the clicked-on will briefly be bolded, Doc Switcher window will close, and then the clicked-entry's tab will become the active tab.

alankilborn:
Generally, I don't like the "selected" terminology, along with parenthetical explanation of it pretty much each time.
Would it be better to call it "pre-selected" (that's what it truly is) and define it once at the top, and then use that term over and over later (without the explanation) each time?

content/docs/user-interface.md
pryrt: Confirrmed and updated

content/docs/user-interface.md
pryrt: Confirmed and updated

pryrt:
> Generally, I don't like the ... parenthetical explanation of it pretty much each time.

I didn't like the parenthetical either.  Your original description focused on the bolding, so I wasn't sure if you thought there would be ambiguity to the reader, so I left it there, though I shouldn't've.

> I don't like the "selected" terminology

I assume that's because it implies to you that the referenced tab is actually "selected", with the follow-on incorrect implication that the tab was immediately activated.  I was trying to draw a distinction between "selecting" in the Document Switcher and the final "activation" of the "selected" tab when the Document Switcher was terminated.   (For a similar distinguishing between "selected" and the final action, in the Plugins Admin interface, you can "select" a Plugin without installing it, and even select it without checkmarking it to put in the queue for installation -- so making a selection and using that selection are two separate things in GUI controls.)

But I can definitely see why you don't like it, because I am sure you aren't the only user who would interpret "select" that way, despite my pedantic usage.

> Would it be better to call it "pre-selected"

I like that even less: I think there are fewer users who would understand that correctly -- I think the first interpretation of that phrase would likely be that it would be a term reserved for whichever tab name is in bold when the popup first opens, and continued use throughout the description would serve to confuse, rather than properly describe.

I am now seeing why you just went with "bold" in your original description.  If I can phrase it in a way I like, I'll probably go back to that idea (or something similar).

alankilborn (approved):

content/docs/user-interface.md
```diff
@@ -116,6 +116,6 @@ When cycling through the tabs in the Document Switcher, if the bottom entry in t
...
+The Document Switcher functionality can also be achieved using just the mouse (if you have a scroll wheel): Right-click in the editing area for a tab and hold the right mouse button, then begin scrolling the mouse wheel (in either direction) to display the Document Switcher popup; further scrolling of the scroll wheel will change which tab is selected in the list. Releasing the right mouse button will cause the currently-selected tab to be activated. An alternate way to activate a tab using the mouse, while the right mouse button is still held and the Document Switcher is displayd, is to left-click on one of the tab names from the list, which will immediately make that entry bold, activate the tab, and close the popup (even though you haven't let go of the right mouse button yet).
```
alankilborn: typo: displayd

PR #635, opened 2024-02-12
Auto-Insertion: restrictions

pryrt:
Describe the safety-restrictions in auto-insertion which prevents unwanted doubles, such as `don''t`


pryrt:
@alankilborn, does this seem a reasonable description?

PR #646, opened 2024-03-06
Toolbar Updates

pryrt:
Describe the **¶** button in the **View > Show Symbol > Show All Characters** description.

In User Interface > Toolbar, make a list of all the built-in toolbar buttons, and link to the menu commands that they activate.


alankilborn:
👍

PR #669, opened 2024-05-03
DirectWrite Improvements

### LINKED ISSUE #668: Clean up DirectWrite mention on in RTL text section
https://npp-user-manual.org/docs/views/#text-direction has two minor issues

1. The phrase "Use Direct Write option" should be "Use DirectWrite option".
2. The current text also is "and if you have that option checkmarked" which should be "and if that option is checkmarked".

The reasoning for the second change is that DirectWrite is enabled by default for new installations since npp v8.6. Thus it may well be enabled for some people even though they did not set the option and to use RTL support they will need to turn it off.

There's also an issue on the https://npp-user-manual.org/docs/preferences/ which still has "This is not on by default, because it may cause performance issues." I have started a [npp forum thread](https://community.notepad-plus-plus.org/topic/25747/directwrite-enabled-by-default-since-v8-6) about that topic.

pryrt:
- Fix "Direct Write" -> "DirectWrite"
- Indicate that before v8.6, "off" was default; but v6.8 and newer, "on" is default
- Reword performance warning closes #668


molsonkiko:
I would change the first sentence to 
* Before v8.6, this *was not* checkmarked by default. Since v8.6, it *is* checkmarked by default.

I think that is a bit clearer.

alankilborn:
I would go a bit further with it, perhaps:

* Before v8.6, this was not checkmarked by default. Since v8.6, it is checkmarked (DirectWrite is "on") by default.

pryrt:
@molsonkiko , @alankilborn , do my updates look sufficient?

molsonkiko:
I approve. It's as clear and concise as could be reasonably expected IMO

alankilborn:
Yep, looks good.

PR #694, opened 2024-08-09
Update preferences.md

TBugReporter:
Corrected some spelling errors


TBugReporter:
Are you using British English or American English for the manual?  I understand that it's best to match whatever Don uses in program prompts, but do you have a policy for the descriptive commentary?

pryrt:
> Are you using British English or American English for the manual? ... do you have a policy for the descriptive commentary?

The goal is to use the spelling that Don uses in the GUI (which is inconsistent, even for the word "colour" in some places vs "color" in others).  The commentary tries to match the spelling of the closest usage when there's conflict, though I don't guarantee it always does.

For example, on the Print options, which you corrected to match the "Color" in the most recent Notepad++ -- that spelling had changed in the GUI, because in v8.0 (and earlier, so in the v7.x when that section of the manual was written), it was still spelled "Colour", which is why the user manual used that spelling.  (I didn't look to see exactly which version changed it; I just had v8.0 conveniently, and since it showed the old spelling, it shows that the GUI changes spelling over time, and we don't always notice.)

PR #709, opened 2024-10-24
Plugin>Notification: add caveat in the "Notification Key" 

pryrt:
need to warn `NPPN_*` notification users that BufferID might not be the active file, and they need to code accordingly.

For now, it's just in the Notification Key, but before merging the PR, will probably need to propagate that caveat to every notification that uses BufferID, and maybe even add a footnote that gives a good sequence for "get active file + bufferID; if argument bufferID != active.bufferID, save the old_active{bufferID,view,index}, switch to view+index for bufferID, perform actions, and optionally switch back to old_active{bufferID,view,index}"  (unfortunately, PythonScript's activateBufferID wraps around multiple calls, so it's more complicated than the [pythonscript example](https://github.com/bruderstein/PythonScript/issues/350#issuecomment-2429015033))

(per conversation with @alankilborn, after he was dealing with [PythonScript SaveAll](https://github.com/bruderstein/PythonScript/issues/350))

### DIFF
--- a/content/docs/plugin-communication.md
+++ b/content/docs/plugin-communication.md
@@ -2307,7 +2307,7 @@ The general layout of the following notifications look like this
 - **Description** informs about the usage of the notification and provides additional information if needed.
 - **Fields** are the parameters to be provided by the notification.
     - ***hwndFrom*** normally holds the **hwndNpp**, which means that the window handle for the current Notepad++ window is passed as that argument.  If it is shown as a `0` or `NULL`, then that notification does not use this Field.  If it is something else, a full description will be provided.
-    - ***idFrom*** normally holds the **BufferID**, which means that the buffer identification integer for the editor buffer of the relevant file is passed as that argument. (_Warning_: this BufferID is often not the _active_ buffer, and if you want to perform an action on it that requires that it be active, you will have to activate that file first; a reasonable sequence for handling this is described in [Notification BufferID](#notification-bufferid).)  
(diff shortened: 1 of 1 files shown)


pryrt:
@alankilborn, per additional chat, I ended up _not_ putting warnings in _all_ the notifications that use BufferID; instead, I have a first warning in the Key; I have a paragraph in the NPPN_FILEBEFORESAVE, and I have a new section at the end which gives details on how to work around the BufferID-vs-active problem.

Does this seem sufficient to you?

content/docs/plugin-communication.md
```diff
@@ -2669,3 +2672,17 @@ The general layout of the following notifications look like this
...
+
+1. Store the BufferID of the active file (`keepBufferID`)
+    - [NPPM_GETCURRENTBUFFERID](#2084nppm_getcurrentbufferid) can be used to populate `keepBufferID`
+2. If the notification BufferID is not the `keepBufferID`, then activate the BufferID.
+	- [NPPM_GETPOSFROMBUFFERID](#2081nppm_getposfrombufferid)`(BufferID)` extracts a value p for the notification BufferID.
+	- Set `v = p >> 30`; set `i = p & 0x3FFFFFFF`.
+	- [NPPM_ACTIVATEDOC](#2052nppm_activatedoc)`(v,i)` will activate the notification BufferID
+3. Perform your actions on the notification BufferID, which is now active.
+4. Set the `keepBufferID` as the active file.
+    - use the same messages as in step 2, but starting with `NPPM_GETPOSFROMBUFFERID(keepBufferID)`
```
alankilborn, line 2688: * Instead of `Set v = p >> 30; set i = p & 0x3FFFFFFF` can we have the better `Set view = p >> 30; set docIndex = p & 0x3FFFFFFF` -- and then later the corresponding `(view, docIndex)` ?
* In step 1, how about adding at the end: `e.g. set keepBufferID = [NPPM_GETCURRENTBUFFERID](...)()

PR #713, opened 2024-11-08
Describe the internal switch from Normal to Extended search mode

pryrt:
cf: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11847


pryrt:
@alankilborn, is this new section sufficient?  Am I missing anything, or did I get the explanation wrong?

alankilborn:
Good description of poor behavior!

PR #726, opened 2024-11-23
Clarify NPPN_FILECLOSED and _FILEBEFORECLOSE

### LINKED ISSUE #719: Improve documentation for close-related plugin messages
Comment: The docs for plugin communication messages in general is "not great". One could spend a lot of time improving this section of the UM wholesale, but maybe this is not a great use of time. Perhaps better is if the section is improved a bit at a time.
I wanted to open this issue as a starting point for discussion.  If it is generally liked, I could certainly propose exact new documentation text.

---

The scenario: Recently I was scripting something that I wanted to have happen when a file/tab is closed.

My starting point was [NPPN_FILECLOSED](https://npp-user-manual.org/docs/plugin-communication/#1005nppn_fileclosed) which currently states: _To notify plugins that the current file is about to be closed_.
Problems with that:
* It isn't necessarily a true statement (consider the case of a _Close All_)
* I was hoping to get the name/path of the file from the notification, all I can get is a `BufferID` -- as documented -- but that id is sort of meaningless in terms of using it to obtain the name/path since the id is invalid after the file/tab is closed.  This is technically not a problem, but it is rather "non obvious"
* In the case of a cloned document, when a _Close_ is issued on the file/tab, the `BufferID` provided _IS_ valid (because one copy of the clone remains)

Because I needed the name/path of the file, I next went to [NPPN_FILEBEFORECLOSE](https://npp-user-manual.org/docs/plugin-communication/#1003nppn_filebeforeclose) which currently states: _To notify plugins that the current file is about to be closed_
Problems:
* Again, it isn't necessarily a true statement (consider the case of a _Close All_)
* (not a true problem but) It might be nice if it were pointed out here that the caller can obtain the name/path of the tab about to close by using the BufferID and calling [NPPM_GETFULLPATHFROMBUFFERID](https://npp-user-manual.org/docs/plugin-communication/#2082nppm_getfullpathfrombufferid) -- which was what I needed to do for my situation
* (not a true problem but) It might be reasonable to point out that this notification seems to exist so that the user can do something while there is a valid BufferID, and that by the time NPPN_FILECLOSED occurs it is (probably) too late

I don't know...maybe I'm being picky.  But I'm always out to improve things that could be improved.  :-)

---

Here's a PythonScript3 one-liner that I found useful during experimentation.  Copy the line and paste into PS3's Console window `>>>` prompt and press Enter:

```py
from Npp import *; notepad.callback(lambda args: print(f'FILEBEFORECLOSE:\n {args=}\n {notepad.getCurrentFilename()=}\n {notepad.getBufferFilename(args["bufferID"])=}\n {notepad.getBufferFilename(notepad.getCurrentBufferID())=}'), [NOTIFICATION.FILEBEFORECLOSE]); notepad.callback(lambda args: print(f'FILECLOSED:\n {args=}\n {notepad.getCurrentFilename()=}\n {notepad.getBufferFilename(args["bufferID"])=}\n {notepad.getBufferFilename(notepad.getCurrentBufferID())=}'), [NOTIFICATION.FILECLOSED])
```

This will generate console output such as the following when file/tabs are closed:

```txt
FILEBEFORECLOSE:
 args={'code': 1003, 'bufferID': 2251062323488}
 notepad.getCurrentFilename()='new 2'
 notepad.getBufferFilename(args["bufferID"])='new 2'
 notepad.getBufferFilename(notepad.getCurrentBufferID())='new 2'
FILECLOSED:
 args={'code': 1005, 'bufferID': 2251062323488}
 notepad.getCurrentFilename()='new 1'
 notepad.getBufferFilename(args["bufferID"])=''
 notepad.getBufferFilename(notepad.getCurrentBufferID())='new 1'
```

Note that in the example, for the `FILECLOSED` output, the `bufferID` of the closed file is invalid causing ` notepad.getBufferFilename(args["bufferID"])` to return an empty string.

pryrt:
closes #719

### DIFF
--- a/content/docs/plugin-communication.md
+++ b/content/docs/plugin-communication.md
@@ -2398,13 +2398,24 @@ The general layout of the following notifications look like this
 ---
 
 #### [1003] **NPPN_FILEBEFORECLOSE**
-*To notify plugins that the current file is about to be closed*
+*To notify plugins that the file referenced by the BufferID is about to be closed.*
 
 **Fields:**
 
 	code:		NPPN_FILEBEFORECLOSE
 	hwndFrom:	hwndNpp
 	idFrom:		BufferID
+
+_Usage Note_: It is useful to store the name of the file that's about to be closed, which can be obtained
+using [NPPM_GETFULLPATHFROMBUFFERID](#2082nppm_getfullpathfrombufferid); plugins need to access this
+information now, rather than while handling [NPPN_FILECLOSED](#1005nppn_fileclosed), because the
+`BufferID`-to-FullPathName mapping will no longer be valid by the time `NPPN_FILECLOSED` is processed.
+
+Also remember, the `BufferID` is not necessarily for the active file; **Close All**, or clicking the close
+button on an inactive tab, or other such file-closing actions, can easily trigger the non-active file to be
+closed; if you need to issue one or more messages that work on the active file, you will need to follow the
+directions in the [Notification BufferID](#notification-bufferid) section (below) while doing the processing
+on the file that's about to be closed.
 
 ---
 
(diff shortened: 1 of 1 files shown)


pryrt:
@alankilborn, since you never provided suggested text in #719, I took a stab at it.  Please review.

alankilborn:
It was still on my RoundTuit list.  :-(

alankilborn:
It looks good.  I compared the content with the sketchy notes I had and everything seems to be covered.

One thing I hadn't fully investigated was closure upon N++ shutdown, to see if there were any special notes needed for that situation.  N++ provides the notifications and closes the files in the active session at shutdown, and I don't think it bothers to do activate-close-activate-close-etc. first, but just does the close, so that situation also falls into the "can't count on it being the active file that is getting closed".

PR #749, opened 2025-01-01
PluginCommunication: Footnotes for Arrays of Path Strings

### LINKED ISSUE #748: NPPM_GETOPENFILENAMES (and similar) Memory Allocation
Per [this Forum Topic](https://community.notepad-plus-plus.org/topic/26509/intermittent-notepad-crash-related-to-nppm_getopenfilenames-only-on-64-bit-npp) and the resulting [issue](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/15997), need to clarify User Manual on the `NPPM_GETOPENFILENAMES` (and similar messages) which want an array of N strings, each [long enough to hold a path](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file "MS Learn: MAX_PATH").

pryrt:
Adds a description of the memory requirements for Arrays of Path Strings, and links to that from the `wchar_t **` messages.
closes #748 

(Also fix from the old `TCHAR` to `wchar_t` throughout the PluginCommunication, to match the recent `Notepad_plus_msgs.h`)

### DIFF
--- a/content/docs/plugin-communication.md
+++ b/content/docs/plugin-communication.md
@@ -1019,7 +1019,7 @@ Puts that mode in the output object. (Added v8.3.3)*
 **Parameters**:
 
 *wParam [in]*
-: const tChar\* languageName, the name of the language to get
+: const wchar_t\* languageName, the name of the language to get
 
 *lParam [out]*
 : ExternalLexerAutoIndentMode &autoIndentMode, an object of the enumeration class ExternalLexerAutoIndentMode, with values:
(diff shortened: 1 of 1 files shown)


pryrt:
@alankilborn and @molsonkiko, how does this look?  Each of the four `wchar_t **` messages links to the footnote, and the footnote explains to use `MAX_PATH` wide characters, which is `MAX_PATH*2` bytes.

alankilborn:
Looks reasonable to me.

molsonkiko:
@pryrt Yep, that looks like a reasonable change to me. Thanks!

PR #793, opened 2025-05-10
Update pt.yaml

Blackspirits:


pryrt:
I will merge the change.  But note that the upstream repository (for the hugo-book theme which is used in the manual) still has the `Buscar`, so if we ever upgrade the theme again, the change could be overwritten.  You might want to go to their repo and put in an issue to ask them to change it.

PR #813, opened 2025-07-04
v8.8.3: self-signed Root CA instructions

pryrt:


pryrt:
confirmed the HTTPS cert download URL is active, so ready to merge

PR #841, opened 2025-11-22
Explain that from Notepad++ 8.8.8, ANSI is disabled when Windows is set to Use Unicode UTF-8 for worldwide language support.

Coises:
Add that beginning with Notepad++ 8.8.8, ANSI is disabled when Windows is set to **Use Unicode UTF-8 for worldwide language support.** Explain why this was done, what happens when Notepad++ opens what users think of as an “ANSI” file, how to determine the Windows setting from the Debug Info and where to find the setting in Windows.


Coises:
Feel free to make this more concise if you can figure out how to do that. It feels long and wordy to me, and like it shouldn’t really be a whole sub-sub-sub-section of its own, but nothing else I tried was any better.

content/docs/preferences.md
```diff
@@ -851,6 +851,16 @@ The entries above the separator line (without `Convert to` in the name) show the
...
+
+Reports suggest that some new installations of Windows 11 enable **Use Unicode UTF-8 for worldwide language support** by default. In Notepad++ you can check the state of this setting by selecting **Debug Info...** from the **?** menu. If the line `Current ANSI codepage` shows `65001` then **Use Unicode UTF-8 for worldwide language support** is enabled; otherwise, it is not.
+
+When Notepad++ loads a file that is recognizably not UTF-8 or UTF-16, it will try to guess the encoding if [MISC > Autodetect character encoding](#misc) is enabled; if autodetection is not enabled or does not yield a positive result, it will make an assumption based on the system locale. Notepad++ then assigns an appropriate encoding from the `Character sets` sub-menus.
+
+**Use Unicode UTF-8 for worldwide language support** is found in Windows Control Panel under **Region** on the **Administrative** tab by clicking the **Change system locale...** button. From Windows **Settings**, you can get there by going to **Time & language** > **Language** or **Language & region** > **Administrative language settings** > **Change system locale**.
```
donho: This MS Windows settings information does not belongs in the npp-user-manual, IMO.

Coises: By way of explanation, I wanted to link the phrase “Use Unicode UTF-8 for worldwide language support” to a Microsoft page that would explain the setting; but I could find no such documentation, aside from [this page](https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page), which is intended for developers and would be confusing to end users. I added that last paragraph only because I could find nothing to which to link that would define what setting was meant. It seems like just giving the text of the setting, without saying where it is found, doesn’t really designate what is meant in a useful way.

pryrt: Like @Coises , I think it's important enough to show a Notepad++ user where they could change it if a Windows setting is messing up their Notepad++ behavior -- especially since someone who was installing Notepad++ on a new PC that had that beta feature enabled by default would be highly confused that Notepad++ was behaving differently on the new computer compared to their expected behavior from the old computer.

pryrt:
> @pryrt What's your opinion?

After getting feedback from my two comments above (ie, assuming you agreed that creating a new Encoding section outside of the preferences page was a good idea), I would hijack this PR (or cancel this one and start my own) to do a bigger rework to create a new section for detailed Encoding documentation.

(I've previously created new sections in the manual without "permission", when I thought it was best for the Manual.  But since you're involved in this discussion now, I want to make sure we're on the same page before I move forward.)

donho:
@pryrt 
OK, It's your PR now.

pryrt:
I've split out most of the Encoding docs into the new `encoding.md`,
I also clarified that the **ANSI** setting isn't always 8-bit, as per @Coises's [side note](#discussion_r2554210460)

@Coises, please make sure I didn't miss anything that was discussed earlier, or mis-explain things in any of my rewording of your text.  Thanks.

pryrt:
@Coises,

Okay, I moved the "if option" to only apply to the heuristic portion, and made sure my order follows yours, without trying to get too far into the nitty-gritty details.  I also made the brief comment about internal representation.

Coises:
> I removed your second sentence and replaced it with a parenthetical showing when it does matter (plugins like HexEdit plugin get confused by the internal representation, though I didn't call out any plugin by name) -- but I've now switched it to mostly your paragraph.

That’s good. I like it.

I mucked up the link for “status bar” and you copied my mistake. It’s in a different file, so I guess it has to be something like `../user-interface/index.html#status-bar` — however that gets done in markdown+Hugo. Sorry about that; I wrote without testing.

Coises:
@pryrt:
> I'll probably let things sit, and look it all over again tomorrow. If I don't see anything else, and you have no other comments, I'll probably publish tomorrow.

I think I’ve run out of things to complain about. ;-)

Thanks for all your work on this — I think it will be much more helpful to users now than my initial changes would have been.

PR #854, opened 2026-03-02
Add "Invalid Regular Expression: Complexity" section

pryrt:
…to help with that common error message


pryrt:
Time and callstack are both resources, IMO.  But yes, calling it "resources" isn't really the best phraseology. :-)

I've reworked that part of the paragraph, to focus on recursion/iteration.  Does it seem more accurate now?

PR #873, opened 2026-05-04
fix fontStyle numbers

CennoxX:
The user manual states italic is 1 and bold 2, but it's the other way around. I found that out, when editing an udl. See the source code of NPP here:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/5ca5caaa6b55a29f9e5270be784fb8316f5bacd6/PowerEditor/src/MISC/Common/NppConstants.h#L299-L305


pryrt:
I thought I had tested that before I put it in the manual, but I guess I messed up my tests back then.  Thanks. :-)

PR #897, opened 2026-08-17
Clarified `\g` and `\k` backreferences in Boost regex.

### LINKED ISSUE #896: [SEARCHING] Difference between `\g` and `\k` is not addressed.
I just spent an hour delving into the possible differences between `\g` and `\k`, of which there are no satisfactory answers, and a few wrong answers. This eventually led me to digging into Boost's [source code](https://github.com/boostorg/regex/tree/develop/include/boost/regex/v5).

The short explanation is that ***there is no difference between `\g` and `\k`*** other than the history behind each.

Boost's source code treats them identically: [regex_traits_defaults.hpp](https://github.com/boostorg/regex/blob/develop/include/boost/regex/v5/regex_traits_defaults.hpp#L459) maps both characters to the exact same internal token:

```c++
escape_type_extended_backref, /*g*/
...
escape_type_extended_backref, /*k*/
```

...and [basic_regex_parser.hpp](https://github.com/boostorg/regex/blob/develop/include/boost/regex/v5/basic_regex_parser.hpp#L856) has one single case for `escape_type_extended_backref`. It never checks if the original character was `g` or `k`.

After digging a little more, it turns out it's just a historical choice based on Perl's implementation that Boost chose to extend.

**For this reason, I think it would be worthwhile mentioning this in the documentation.**

My proposed change would be to add the following under the [Capture Groups and Backreferences](https://npp-user-manual.org/docs/searching/#capture-groups-and-backreferences) section of the `Searching` page:

> * `\g` and `\k` are functionally synonymous prefixes for _extended backreferences._ There is no semantic difference in Boost's implementation. They coexist for syntax-compatibility/historical reasons.

I'm about to make a PR for this change.

pbarney:
According to Boost's source code, `\g` and `\k` backreferences are synonymous. This edit removes any ambiguity that may remain in the reader's mind regarding when to use one or the other.

Fixes #896

### DIFF
--- a/content/docs/searching.md
+++ b/content/docs/searching.md
@@ -747,8 +747,6 @@ Anchors match a zero-length position in the line, rather than a particular chara
 
 * `(?<name>`_subset_`)` or `(?'name'`_subset_`)` ⇒ _Named Capture Group_: Names the value matched by _subset_ as the group _name_.  Please note that group names are case-sensitive.
 
-* `\g` and `\k` are functionally synonymous prefixes for _extended backreferences._ There is no semantic difference in Boost's implementation. They coexist for syntax-compatibility/historical reasons.
-
 * `\ℕ`, `\gℕ`, `\g{ℕ}`, `\g<ℕ>`, `\g'ℕ'`, `\kℕ`, `\k{ℕ}`, `\k<ℕ>` or `\k'ℕ'` ⇒ _Numbered Backreference:_ These syntaxes match the ℕth capture group earlier in the same expression.  (Backreferences are used to refer to the capture group contents only in the search/match expression; see the [Substitution Escape Sequences](#substitution-escape-sequences) for how to refer to capture groups in substitutions/replacements.)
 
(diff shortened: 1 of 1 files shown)


pryrt:
If you must clarify these, then do it as a subpoint to the already existing bullet point, don't put your "caveat" before those symbols are defined.

content/docs/searching.md
```diff
@@ -747,6 +747,8 @@ Anchors match a zero-length position in the line, rather than a particular chara
 
 * `(?<name>`_subset_`)` or `(?'name'`_subset_`)` ⇒ _Named Capture Group_: Names the value matched by _subset_ as the group _name_.  Please note that group names are case-sensitive.
 
+* `\g` and `\k` are functionally synonymous prefixes for _extended backreferences._ There is no semantic difference in Boost's implementation. They coexist for syntax-compatibility/historical reasons.
```
pryrt: Needs to be moved down...  I would say move it after the `For negative ℕ` a few paragraphs down, and make it have the same indentation as that note.

pbarney: Updated. Please let me know if I misunderstood your placement request.

PR #898, opened 2026-08-19
Revise Plugins Admin buttons for installed, incompatible, and deactivated plugins

bms-2015:
Updated the Plugins Admin section to reflect changes in button functionality for the Installed, Incompatible, and Deactivated tabs. Installed plugins now provide both Deactivate and Remove buttons. Incompatible plugins now provide a Remove button. Deactivated plugins now provide Activate and Remove buttons. These updates ensure that the documentation accurately reflects the current behavior of the Notepad++ interface.

### DIFF
--- a/content/docs/plugins.md
+++ b/content/docs/plugins.md
@@ -85,10 +85,10 @@ The Plugins Admin has five tabs:
 
 - **Available** ⇒ Shows plugins you haven't yet installed, and gives an **Install** button
 - **Updates** ⇒ Shows plugins you have installed but that Plugins Admin knows have updates available, and gives an **Update** button
-- **Installed** ⇒ Shows plugins you have installed, and gives a **Remove** button
-- **Incompatible** ⇒ Shows plugins you had previously installed but are no longer compatible with Notepad++:
+- **Installed** ⇒ Shows plugins you have installed, and gives **Deactivate** and **Remove** buttons
+- **Incompatible** ⇒ Shows plugins you had previously installed but are no longer compatible with Notepad++, and gives a **Remove** button:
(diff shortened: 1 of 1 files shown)


bms-2015:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/18206#issue-4870937404

content/docs/plugins.md
```diff
@@ -105,7 +105,7 @@ Actions available through the Plugins
     - This button is visible when you are in the **Updates** tab.
     - Checkmark the plugin(s) you want to deactivate, and click **Update**.
         - Notepad++ will install the more-recent version of the plugin, then restart.
-- **Deactivate**: Allows you to deactivate one or more plugins using the GUI.
+- **Deactivate**: Allows you to deactivate one or more plugins using the GUI. Deactivated plugins remain in the program’s folder but are not loaded at startup, thereby reducing memory usage and decluttering the interface.
```
pryrt: Doesn't this contradict the actual behavior (or at least muddle the issue), where it moves the plugin into a new subdirectory (the `disabled\` subdirectory), as is explicitly described below?

When I read "remain in the program's folder", that implies that they remain unmoved to me.  But in actuality, it does move where it is, within the program's folder hierarchy.

I would rephrase your new text as:
> ... Deactivated plugins are moved to a new subdirectory (see below), and are thus not loaded at startup, thereby reducing...

bms-2015: Okay, I've changed the item. They're still in the program folder (NPP), but It doesn't really matter. It was more important to me to demonstrate the purpose of deactivation.

PR #907, opened 2026-09-13
Add Session Network Security section to session.md

pryrt:
per https://github.com/notepad-plus-plus/notepad-plus-plus/issues/18319 and https://github.com/notepad-plus-plus/notepad-plus-plus/pull/18374


content/docs/session.md
```diff
@@ -73,6 +73,40 @@ Some people want to be able to open a new instance of Notepad++ when they double
...
+- **Always skip network paths** if you never want to load network paths from session files
+- **Always load network paths** if you want to always load any network paths from session files
+
+It should remember your "always" choice for this instance of Notepad++, and will save your choice to `<NetworkPathsAlwaysAction value="#" />` in `serverWhiteList.xml` (starting in v8.9.8.1) :
+
+- `value="?"`: will prompt next time
+- `value="!"`: will always skip loading network files from session files
+- `value="*"`: will always load network files from session files
+- Notepad++ saves this file as the application exits (in multi-instance mode, your instance must be the one that has permission to save the configuration), so if you launch another instance before you've exited, and that instance tries to load a UNC path from the session, you will be prompted again (because newly launched instances do not inherit changed-but-not-yet-saved configuration settings).
```
xomx: @pryrt 
N++ updates the serverWhiteList.xml content on disk `immediately`, not as usual at the N++ exit, right after the user clicks on one of the new NetPathWarning dlg buttons. Fortunately, a user cannot click in N++ multi-instance mode on more than one NetPathWarning dlg button at once, otherwise there can easily be a race condition in the SWL saving.

It's a must here, otherwise there could be a discrepancy in between the currently running N++ instance internal setting and the SWL file content (which gets always loaded when a user uses e.g. the NppShell context menu for opening of another file in the already running N++; or by any other way indirectly launching a new instance of N++).

pryrt: @xomx,

If INSTANCE1 saves the file with adding `MyServer`, will INSTANCE2 re-read the file before deciding whether to prompt again, or will it again ask about `MyServer` even though it's already been added by the other instance?

xomx: Good point.

There are two(three) types of INSTANCE2.

1) INSTANCE2 in N++ default mono-instace mode.
In single-instance mode, when INSTANCE1 is already running, the INSTANCE2 is created e.g. by the NppShell "Edit with Notepad++" context menu click (WinOS itself will launch that INSTANCE2, with the right-clicked files to be opened in N++ as its cmdline param). The same is valid for any cmdline notepad++.exe launching. Since such a INSTANCE2 is starting from scratch, it will always load & have current content of the SWL file. So it will know about your "MyServer" record.

2) INSTANCE2 in N++ multi-instace mode.
I'm afraid that I forgot to take this into account (such a minor thing that NetPathWarnDlg and it gets pretty complicated, _sigh_). In this case, two scenarios:
- a) you just launched that INSTANCE2 - then everything is ok (as this is the same as the case 1)
- b) INSTANCE2 is already running - then there is a catch, if the "MyServer" was added from the INSTANCE1 into SWL file after the INSTANCE2 launch, the INSTANCE2 will not know about that "MyServer"

---

Why it is a problem in the 2b and what will be the practical outcome now:
- currently, if the SWL file is found at the very N++ instance start, its content is loaded (& checked for the correct format - if corrupted, it gets reset to safe default - the "?" mode and no allowed server items)
- then, when a user clicks the buttons in that NetPathWarnDlg, the selected choice (Skip/Load(WhiteListServerName)/AlwayLoad/AlwaysSkip) is stored into internal N++ in-memory xml storage and then also immediately reflected/saved into the SWL file (updated an existing or created a new one)
- so the INSTANCE2 internal in-memory xml storage may become different from the INSTANCE1 one, depending on which net-files the user opens/approves in the respective instance
- in the SWL file then can be duplicated allowed server names, as the individual instances decide what to write to SWL file based on their internal XML data and not on content of the SWL file
- as for the SWL mode ?/!/* saved - the last change made by the individual instances wins (which is ok, I'd expect that from the user perspective)

So, the worst that should happen now is that a user of multi-instance mode might occasionally have to allow or deny the same server multiple times (depending on the number of running instances), and that duplicate entries for allowed servers might accumulate in the SWL file.

I can try to fix that later (no earlier than Friday).

